PR #2, opened 2020-04-30
WIP record parser

uranusjr:
Let’s get something started? :p

First posted to make sure I’m on the right track. ~I’ll rewrite this to be Python 2.7 compatible if this approach looks valid.~ Already did.


src/installer/wheels.py
```diff
@@ -0,0 +1,52 @@
...
+    @classmethod
+    def parse(cls, s: str) -> Optional[Hash]:
+        name, value = s.split("=", 1)
+        return cls(name, value)
+
+
+@dataclasses.dataclass()
+class Record:
+    path: pathlib.PurePosixPath
+    hash_: Optional[Hash]
```
brettcannon, line 29: Does this need to be its own dataclass, or could it just be a tuple?

tests/test_wheels.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_parse_wheel_record_simple():
+    record_content = textwrap.dedent(
+        """\
+        file.py,sha256=AVTFPZpEKzuHr7OvQZmhaU3LvwKz06AJw8mT\\_pNh2yI,3144
+        distribution-1.0.dist-info/RECORD,,
+        """
+    )
+    records = list(parse_record_file(io.StringIO(record_content)))
```
brettcannon, line 16: If the function is going to take an iterator of strings, then the example text can just be the `iter()` of a list of lines. That might also make it easier to work with from a `pytest.mark.parametrize()` perspective.

tests/test_wheels.py
```diff
@@ -0,0 +1,88 @@
...
+    record_content = textwrap.dedent(
+        """\
+        file.py,sha256=AVTFPZpEKzuHr7OvQZmhaU3LvwKz06AJw8mT\\_pNh2yI,3144,
+        distribution-1.0.dist-info/RECORD,,
+        """
+    )
+    record_io = io.StringIO(record_content)
+
+    with pytest.warns(SuperfulousRecordColumnsWarning) as ws:
+        records = list(parse_record_file(record_io))
```
brettcannon, line 41: Is the expectation to warning once in total, or once per extraneous column? This test doesn't suggest which is expected.

tests/test_wheels.py
```diff
@@ -0,0 +1,88 @@
...
+
+    assert len(ws) == 1
+    assert ws[0].message.args[0] == "Dropping columns [3:] from row 0"
```
brettcannon, line 44: I wouldn't test the exact message as I find that to typically be brittle. I also don't think the message should be part of the API. At best I would check that `"0"` is in the message if you want to guarantee the troublesome row is mentioned.

tests/test_wheels.py
```diff
@@ -0,0 +1,88 @@
...
+            RECORD_CONTENT_INVALID_SIZE,
+            ["distribution-1.0.dist-info/RECORD", "", "deadbeef"],
+        ),
+    ],
+)
+def test_parse_wheel_record_invalid(record_content, invalid_row):
+    record_io = io.StringIO(record_content)
+    with pytest.raises(ValueError) as ctx:
+        list(parse_record_file(record_io))
+    assert str(ctx.value) == f"invalid row 1: {invalid_row!r}"
```
brettcannon, line 88: Would it be worth having a test that read from an actual file since that will be a common case use? It could also test against a real-world `RECORD` file.

src/installer/wheels.py
uranusjr: I think it will be a `collections.namedtuple()` eventually (considering the plan seems to be to support Python 2)

tests/test_wheels.py
uranusjr: Good point, I’ll add more errornous rows to the input. The expectation is to warn once for each row (for no particular reasons beyond it’s what pip does right now and easy to write.)

tests/test_wheels.py
uranusjr: Think about this a bit further, maybe it’s wrong to accept an iterable as argument, because this precludes the possibility to use this in an async generator. Maybe it should work like this instead:

```python
def get_record_file_parser():
    ...

parser = get_record_file_parser()
for line in lines:  # Can be async for as well.
   record = parser.send(line)
   # Do something with record...
```

And then we can have a synchronous wrapper of this to accept a synchronous interable.

uranusjr: Scratch that, `csv.reader()` is not line-by-line, so `send()` wouldn’t work.

tests/test_wheels.py
uranusjr: Added to `test_parse_wheel_record_simple`.

noxfile.py
```diff
@@ -27,6 +27,7 @@ def test(session):
         "pytest",
         "--cov=installer",
         "--cov-fail-under=100",
+        "--cov-report=term-missing",
```
gaborbernat, line 30: 😭 nox over tox

uranusjr: That’s @pradyunsg, I just use whatever is available 🤷

uranusjr:
Any suggestion on the module/package name? I need somewhere to put the parser, and `wheels.py` is rather arbitrary. Another idea is `records.py` which can also contain the RECORD *writer*.

gaborbernat:
I agree on ``records.py`` that's part of a package ``installer``. Generally, the installer will only understand wheels so I don't think it makes sense to add a ``wheels`` package.

src/installer/wheels.py
uranusjr: Thinking about this a bit more, maybe Hash really needs to be a class, so we can put a `validate(data: bytes) -> bool` on it.

src/installer/records.py
```diff
@@ -0,0 +1,59 @@
...
+        if len(row) > 3:
+            warnings.warn(
+                "Dropping columns [3:] from row {}".format(i),
+                SuperfulousRecordColumnsWarning,
+            )
+        try:
+            record = Record.parse(row[0], row[1], row[2])
+        except (IndexError, ValueError):
+            raise ValueError("invalid row {}: {!r}".format(i, row))
+        yield record
```
gaborbernat, line 59: Can you define inside ``__all__`` methods meant to be used from outside?

uranusjr: Done!

src/installer/records.py
```diff
@@ -0,0 +1,73 @@
...
+        return cls(
+            path=pathlib.PurePosixPath(p),
+            hash_=Hash.parse(h) if h else None,
+            size=int(s) if s else None,
+        )
+
+
+def parse_record_file(f):
+    # type: (Iterator[str]) -> Iterator[Record]
+    for i, row in enumerate(csv.reader(f)):
```
gaborbernat, line 63: nitpick: I usually name these position arguments ``at_pos`` as feel makes the code easier to read in the body, ``i`` doesn't tell much what it encodes; generally reserve single character variables for comprehension 🤔

src/installer/records.py
```diff
@@ -0,0 +1,73 @@
...
+        if len(row) > 3:
+            warnings.warn(
+                "Dropping columns [3:] from row {}".format(i),
+                SuperfulousRecordColumnsWarning,
+            )
+        try:
+            record = Record.parse(row[0], row[1], row[2])
+        except (IndexError, ValueError):
```
gaborbernat, line 71: what throws here the ValueError?

noxfile.py
gaborbernat: just nitpick, 😆 feel free to ignore; but felt as maintainer of tox must make the comment

gaborbernat, line 30: I usually like to also generate the HTML report with context to ``{envtmpdir}`` 👍

pyproject.toml
```diff
@@ -18,3 +21,6 @@ test = [
   "pytest-xdist",
 ]
 doc = ["sphinx"]
+
+[tool.black]
+line-length = 79
```
gaborbernat, line 26: can we leave this at blacks default?

pradyunsg: +1

I know there's 1 knob in black and... I'd like us to stick with the default and avoid debating that. 🙈

src/installer/records.py
uranusjr: `Record.parse()`. I’ll add a docstring to make this more obvious.

noxfile.py
uranusjr: Do you happen to recall the argument to use? I don’t use that often and it’s a hassle to dig into documentation.

pyproject.toml
uranusjr: I hate Black’s 88 default and feel very strongly it’s interpreting the 80±10% rule wrong. But whatever, I have more important things to worry about.

While we’re on the topic of default values though, @pradyunsg can I move `tools/isort.cfg` to `.isort.cfg` (which is the default settings location)? I don’t use pre-commit, and the custom config location is messing with my manual isort calls :/

uranusjr:
Heh, isort and black are refusing to work together.

noxfile.py
gaborbernat: ``
--cov-report=html:{toxworkdir}/htmlcov \
--cov-context=test \
```

File: pyproject.toml (hunk 10 above)
Comment (gaborbernat) [reply pyproject.toml]: Out of curiosity why not use pre-commit?

File: noxfile.py (hunk 6 above)
Comment (uranusjr) [reply noxfile.py:31]: Thanks!

File: pyproject.toml (hunk 10 above)
Comment (uranusjr) [reply pyproject.toml]: It makes `git commit` extremely slow on Windows.

**Comment by brettcannon** [comment]
@uranusjr we get semi-regular issues from isort and Black users on the Python extension for VS Code from complaining about them competing with each other. 😆

File: pyproject.toml (hunk 10 above)
Comment (gaborbernat) [reply pyproject.toml]: Don't install the auto pre-commit feature? I usually call it manually as in ``tox -e fix_lint`` 😊

**Comment by gaborbernat** [comment]
Did we set up the configuration to match the style? If so should be just one issue (when there's a comment for an import)?

File: .gitignore [hunk 11]
```diff
@@ -3,6 +3,7 @@
 .nox/
 /build/
 /dist/
+/htmlcov/
```
gaborbernat, line 6: I don't like putting it at root, does not allow to run multiple target pythons in parallel; this path needs to be per nox env different

uranusjr: I see, thanks!

pyproject.toml
pradyunsg: > While we’re on the topic of default values though, @pradyunsg can I move `tools/isort.cfg` to `.isort.cfg` (which is the default settings location)?

Sure, go ahead and change that. :)

src/installer/_compat/typing.py
```diff
@@ -0,0 +1,6 @@
+__all__ = ["TYPE_CHECKING"]
+
+try:  # pragma: no cover
+    from typing import TYPE_CHECKING
+except ImportError:  # pragma: no cover
+    TYPE_CHECKING = False
```
pradyunsg, line 6:
```suggestion
if False:
    from typing import TYPE_CHECKING
else:
    TYPE_CHECKING = False
```

nit: let's not import typing even on Python 2 (to avoid uninstalling-in-use-module issues if pip vendors this)?

src/installer/records.py
```diff
@@ -0,0 +1,84 @@
...
+        return "Hash(name={!r}, value={!r})".format(self.name, self.value)
+
+    @classmethod
+    def parse(cls, h):
+        # type: (str) -> Hash
+        name, value = h.split("=", 1)
+        return Hash(name, value)
+
+
+class Record(object):
```
pradyunsg, line 38: nit:

```suggestion
class RecordItem(object):
```

pradyunsg, line 38: nit:

```suggestion
class RecordItem(object):
```

src/installer/_compat/typing.py
pradyunsg, line 6:
```suggestion
if False:
    from typing import TYPE_CHECKING
else:
    TYPE_CHECKING = False
```

nit: let's not import typing even on Python 2 (to avoid uninstalling-in-use-module issues if pip vendors this)?

uranusjr: Why not just `TYPE_CHECKING = False`?

pradyunsg:
:/

Don't merge PRs with a WIP in the heading?

src/installer/_compat/typing.py
pradyunsg: To make typing actually treat this value as "ALWAYS FALSE".

gaborbernat:
I thought draft PRs are for that 🤔

pradyunsg:
I agree that draft PRs also indicate this in a much better way. However, the title clearly says "WIP" (it still does) which is a fairly non-ambiguously indicates the same thing.

If y'all reckon PRs are OK to merge while "WIP" is the first thing in their titles, I disagree with that. Anyway, it was very surprising that a PR with WIP in it's title got merged, and I was certainly caught off guard by that.

I don't really have much to say other than, well, I was caught off guard.

PR #3, opened 2020-05-02
An almost working implementation to install from .whl

uranusjr:
I implemented the whole workflow to identify what are the missing pieces. And it’s almost working!

```console
$ py examples/whl.py <..whl> <directory> [--installer <installer-name>]
```

(See pypa/pip#8156 for background of the installer name feature.)

There are two areas I have not implemented:

* Script generation. This is probably the most hairy part unimplmented. I’ll probably go ~copy~ reference parts of distlib.
* PEP 610 (`direct_url.json`). I’m not sure about how we should implement this. pip already has an implementation, but I don’t feel like copying it directly since we only need a small subset of it (to write the file, not read and make sense of it). Maybe we should put it somewhere instead, e.g. `packaging` or `importlib-metadata`? The installer can just accept an optional `DirectURL` instance and work with that.

And then we can start identifying features we can make sans-IO. I already pulled out a part of making sense of the wheel structure (where the `.dist-info` directory is, and format various files in it). Some other potential parts I can think of:

* Metadata generation (other than `RECORD`). Especially if we have PEP 610 support, since that is more than writing a string to a file.
* Keeping track of the installed files. Currently the example is using a dict to do this, but maybe we can have a Recorder (I’m terrible at naming) class to encapsulate the logic.

### DIFF
--- a/examples/whl.py
+++ b/examples/whl.py
@@ -42,11 +42,11 @@ class ZipFileInstaller(object):
     """Install a local wheel.
     """
 
-    def __init__(self, name, dist_info, zip_file_handler):
+    def __init__(self, name, dist_info, zip_file_handle):
         # type: (str, DistInfo, zipfile.ZipFile) -> None
         self._name = name
         self._dist_info = dist_info
-        self._zip_file_handler = zip_file_handler
+        self._zip_file_handle = zip_file_handle
 
     @classmethod
     @contextlib.contextmanager
@@ -85,7 +85,7 @@ class ZipFileInstaller(object):
 
     def _install_record_item(self, item, directory):
         # type: (RecordItem, pathlib.Path) -> None
-        with self._zip_file_handler.open(str(item.path)) as f:
+        with self._zip_file_handle.open(str(item.path)) as f:
             data = f.read()
         item.raise_for_validation(data)
         target = directory.joinpath(item.path)
@@ -95,7 +95,7 @@ class ZipFileInstaller(object):
 
     def _iter_installed_record_items(self, directory):
         # type: (pathlib.Path) -> Iterator[RecordItem]
-        with self._zip_file_handler.open(str(self._dist_info.record)) as f:
+        with self._zip_file_handle.open(str(self._dist_info.record)) as f:
             for item in parse_record_file(_wrap_as_io_str(f)):
                 self._install_record_item(item, directory)
                 yield item
--- a/examples/whl.py
+++ b/examples/whl.py
@@ -69,6 +69,7 @@ class ZipFileInstaller(object):
         target = directory.joinpath(item.path)
         with self._open_target_for_write(target, binary=True) as f:
             f.write(data)
+        # TODO: Handle file permission and other metadata.
 
     def _iter_installed_record_items(self, directory):
         # type: (pathlib.Path) -> Iterator[RecordItem]
@@ -100,9 +101,11 @@ class ZipFileInstaller(object):
     def install(self, directory):
         # type: (pathlib.Path) -> None
         items = {r.path: r for r in self._iter_installed_record_items(directory)}
+        # TODO: Install .data directory.
         items.update((r.path, r) for r in self._iter_installed_scripts(directory))
         items.update((r.path, r) for r in self._write_additional_metadata(directory))
         self._write_record(directory, items)
+        # TODO: Compile .pyc files.
 
 
 def main(args=None):
--- a/examples/whl.py
+++ b/examples/whl.py
@@ -42,11 +42,11 @@ class ZipFileInstaller(object):
     """Install a local wheel.
     """
 
-    def __init__(self, name, distinfo, zf):
+    def __init__(self, name, distinfo, zip_file_handler):
         # type: (str, DistInfo, zipfile.ZipFile) -> None
         self._name = name
         self._distinfo = distinfo
-        self._zf = zf
+        self._zip_file_handler = zip_file_handler
 
     @classmethod
     @contextlib.contextmanager
@@ -85,7 +85,7 @@ class ZipFileInstaller(object):
 
     def _install_record_item(self, item, directory):
         # type: (RecordItem, pathlib.Path) -> None
-        with self._zf.open(str(item.path)) as f:
+        with self._zip_file_handler.open(str(item.path)) as f:
             data = f.read()
         item.raise_for_validation(data)
         target = directory.joinpath(item.path)
@@ -95,7 +95,7 @@ class ZipFileInstaller(object):
 
     def _iter_installed_record_items(self, directory):
         # type: (pathlib.Path) -> Iterator[RecordItem]
-        with self._zf.open(str(self._distinfo.record)) as f:
+        with self._zip_file_handler.open(str(self._distinfo.record)) as f:
             for item in parse_record_file(_wrap_as_io_str(f)):
                 self._install_record_item(item, directory)
                 yield item
--- a/examples/whl.py
+++ b/examples/whl.py
@@ -42,21 +42,21 @@ class ZipFileInstaller(object):
     """Install a local wheel.
     """
 
-    def __init__(self, name, distinfo, zip_file_handler):
+    def __init__(self, name, dist_info, zip_file_handler):
         # type: (str, DistInfo, zipfile.ZipFile) -> None
         self._name = name
-        self._distinfo = distinfo
+        self._dist_info = dist_info
         self._zip_file_handler = zip_file_handler
 
     @classmethod
     @contextlib.contextmanager
-    def create(cls, name, wheel_path):
+    def from_wheel_path(cls, name, wheel_path):
         # type: (str, pathlib.Path) -> Iterator[ZipFileInstaller]
         project_name, project_version, _ = wheel_path.stem.split("-", 2)
         with zipfile.ZipFile(str(wheel_path)) as zf:
             entry_names = (name.lstrip("/").split("/", 1)[0] for name in zf.namelist())
-            distinfo = DistInfo.find(project_name, project_version, entry_names)
-            yield cls(name, distinfo, zf)
+            dist_info = DistInfo.find(project_name, project_version, entry_names)
+            yield cls(name, dist_info, zf)
 
     @contextlib.contextmanager
     def _open_adjacent_tmp_for_write(self, path, **kwargs):
@@ -95,7 +95,7 @@ class ZipFileInstaller(object):
 
     def _iter_installed_record_items(self, directory):
         # type: (pathlib.Path) -> Iterator[RecordItem]
-        with self._zip_file_handler.open(str(self._distinfo.record)) as f:
+        with self._zip_file_handler.open(str(self._dist_info.record)) as f:
             for item in parse_record_file(_wrap_as_io_str(f)):
                 self._install_record_item(item, directory)
                 yield item
@@ -106,15 +106,15 @@ class ZipFileInstaller(object):
 
     def _write_additional_metadata(self, directory):
         # type: (pathlib.Path) -> Iterator[RecordItem]
-        installer = directory.joinpath(self._distinfo.installer)
+        installer = directory.joinpath(self._dist_info.installer)
         with self._open_target_for_write(installer) as f:
             f.write(self._name)
-        yield RecordItem(self._distinfo.installer, None, None)
+        yield RecordItem(self._dist_info.installer, None, None)
         # TODO: Write direct_url.json.
 
     def _write_record(self, directory, installed_items):
         # type: (pathlib.Path, Dict[pathlib.PurePosixPath, RecordItem]) -> None
-        record = self._distinfo.record
+        record = self._dist_info.record
         installed_items[record] = RecordItem(record, None, None)
         with self._open_csv_for_write(directory.joinpath(record)) as f:
             write_record_file(f, installed_items.values())
--- a/examples/whl.py
+++ b/examples/whl.py
@@ -4,8 +4,12 @@
 import argparse
 import codecs
 import contextlib
+import errno
+import os
 import sys
 import zipfile
+
+import six
 
 from installer._compat import pathlib
 from installer._compat.typing import TYPE_CHECKING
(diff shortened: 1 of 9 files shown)


examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def install(self, directory):
+        # type: (pathlib.Path) -> None
+        items = {r.path: r for r in self._iter_installed_record_items(directory)}
+        items.update((r.path, r) for r in self._iter_installed_scripts(directory))
+        items.update((r.path, r) for r in self._write_additional_metadata(directory))
+        self._write_record(directory, items)
+
+
+def main(args=None):
```
gaborbernat, line 111: Can we please move out all CLI related parts to ``__main__.py`` instead?

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+
+class ZipFileInstaller(object):
+    """Install a local wheel.
+    """
+
+    def __init__(self, name, distinfo, zf):
+        # type: (str, DistInfo, zipfile.ZipFile) -> None
+        self._name = name
+        self._distinfo = distinfo
+        self._zf = zf
```
gaborbernat, line 27: can we rename this to zip_file_handler instead, ``zf`` doesn't read well in my mind 🤔

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @classmethod
+    @contextlib.contextmanager
+    def create(cls, name, wheel_path):
```
gaborbernat, line 31: Wouldn't a ``from_wheel_path`` be a better name for this?

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+        # type: (str, pathlib.Path) -> Iterator[ZipFileInstaller]
+        project_name, project_version, _ = wheel_path.stem.split("-", 2)
+        with zipfile.ZipFile(str(wheel_path)) as zf:
+            entry_names = (name.lstrip("/").split("/", 1)[0] for name in zf.namelist())
+            distinfo = DistInfo.find(project_name, project_version, entry_names)
```
gaborbernat, line 36: nitpick: I'd prefer this to be ``dist_info``... even in the file name we split by ``-`` so don't see why we would not do the same with the variable name, makes it easier to read

src/installer/layouts.py
```diff
@@ -0,0 +1,83 @@
+__all__ = ["DistInfo"]
```
gaborbernat, line 1: why put this at the top, in my experience this is always at the end of file

src/installer/layouts.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @classmethod
+    def find(cls, project_name, project_version, entry_names):
+        # type: (str, str, Iterable[str]) -> DistInfo
+        escaped_project_name = _name_escape(project_name).lower()
+        escaped_project_version = _version_escape(project_version)
+
+        for entry_name in entry_names:
+            stem, ext = os.path.splitext(entry_name)
+            if ext.lower() != ".dist-info":
```
gaborbernat, line 54: consider making the suffix a private class attribute?

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+    def _write_record(self, directory, installed_items):
+        # type: (pathlib.Path, Dict[pathlib.PurePosixPath, RecordItem]) -> None
+        record = self._distinfo.record
+        installed_items[record] = RecordItem(record, None, None)
+        with self._open_csv_for_write(directory.joinpath(record)) as f:
+            write_record_file(f, installed_items.values())
+
+    def install(self, directory):
+        # type: (pathlib.Path) -> None
+        items = {r.path: r for r in self._iter_installed_record_items(directory)}
```
gaborbernat, line 103: Do we need the iter keyword, feels like redundant to me 🤔

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            kwargs = {"mode": "w", "newline": "", "encoding": "utf-8"}
+        return self._open_adjacent_tmp_for_write(path, **kwargs)
+
+    def _install_record_item(self, item, directory):
+        # type: (RecordItem, pathlib.Path) -> None
+        with self._zf.open(str(item.path)) as f:
+            data = f.read()
+        item.raise_for_validation(data)
+        target = directory.joinpath(item.path)
```
gaborbernat, line 69: instead of join path why not just use the ``/`` operator?

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+            yield cls(name, distinfo, zf)
+
+    @contextlib.contextmanager
+    def _open_adjacent_tmp_for_write(self, path, **kwargs):
+        # type: (pathlib.Path, Any) -> Iterator[IO[Any]]
+        path.parent.mkdir(parents=True, exist_ok=True)
+        temp = path.with_name("{}.tmp.{}".format(path.name, self._name))
+        with temp.open(**kwargs) as f:
+            yield f
+        temp.replace(path)
```
gaborbernat, line 46: given it also replaces I feel like just open is not adequate naming here

gaborbernat: then again what's the point of this tmp file rather than just direct-write?

examples/whl.py
```diff
@@ -0,0 +1,120 @@
...
+    parser.add_argument("wheel", type=pathlib.Path)
+    parser.add_argument("dest", type=pathlib.Path)
+    parser.add_argument("--installer", default="pypa-installer")
+
+    options = parser.parse_args(args)
+    with ZipFileInstaller.create(options.installer, options.wheel) as installer:
+        installer.install(options.dest)
+
+
+if __name__ == "__main__":
```
gaborbernat, line 122: yeah this should go into the ``__main__`` 👍

pyproject.toml
```diff
@@ -11,7 +11,8 @@ description-file = "README.md"
 classifiers = ["License :: OSI Approved :: MIT License"]
 requires-python = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
 requires = [
-  "pathlib2; python_version < '3.4'"
+  "pathlib2; python_version < '3.4'",
+  "six",
```
gaborbernat, line 15: considering the size of this IMHO we should try to make it work without ``six``

pyproject.toml
```diff
@@ -11,7 +11,8 @@ description-file = "README.md"
 classifiers = ["License :: OSI Approved :: MIT License"]
 requires-python = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
 requires = [
-  "pathlib2; python_version < '3.4'"
+  "pathlib2; python_version < '3.4'",
```
gaborbernat, line 14: pathlib2 is quite heavyweight, I'd recommend instead of this import lets just make a small ``_pathlib`` file we can fall back to that uses os.path operations underneath, this probably requires just like 5 methods to implement for now

src/installer/exceptions.py
```diff
@@ -0,0 +1,27 @@
...
+    "InvalidWheel",
+    "MetadataNotFound",
+    "RecordItemError",
+    "RecordItemHashMismatch",
+    "RecordItemSizeMismatch",
+]
+
+
+class InvalidWheel(Exception):
+    pass
```
gaborbernat, line 11: instead of pass consider adding a docstring that explains when the error may be thrown

src/installer/layouts.py
gaborbernat: the name of this file is ``layouts`` but it really only handles ``DistInfo`` should we just name it ``dist_info.py`` then?

examples/whl.py
uranusjr: I stole this pattern from pip and thought the original implementation must exists for a reason. Likely some kind of edge case about writing to an already-open file.

I’m not really in love with the naming of this either, but can’t think of a better one.

examples/whl.py
uranusjr: The slash operator is a mistake and should have never existed IMO.

examples/whl.py
uranusjr: I think this (or some kind of a similar implementation based on it) will end up there eventually, but I want to first distil the implementation 🤔

src/installer/layouts.py
uranusjr: Hmm, in mu experience `__all__` is always at the beginning (either at the very top or immediately below imports)

examples/whl.py
gaborbernat: Can't this function be distilled in another file? Doesn't seem to belong in the library part IMHO.

src/installer/layouts.py
gaborbernat: I've never seen it at the beginning... I've certainly put it at the end always and goes well with the lint rule that imports are at the top, constants afterwards.

examples/whl.py
gaborbernat: I strongly disagree. IMO makes it a **lot** more readable, this just reads badly to me.

examples/whl.py
gaborbernat: @pradyunsg care to enlighten us? for me at the moment seems just a needless complication 😊 At least let's document it why we need this awkward way.

examples/whl.py
uranusjr: Not sure I follow. This file is currently not in the library at all.

gaborbernat: Oh wait, it's an example, I've thought it's within the package 😊 nm then.

src/installer/layouts.py
uranusjr: As I said, my experience is otherwise. Django’s coding convention puts it right after imports, for example.

Some other examples for the standard library (not always a good source for coding conventions, but nonetheless):

* [Right after imports](https://github.com/python/cpython/blob/48b069a003ba6c684a9ba78493fbbec5e89f10b8/Lib/os.py). There are a lot of examples.
* [At the very beginning](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Lib/struct.py)

My quick search in stdlib also didn’t find any examples putting `__all__` at the end.

gaborbernat: I'm agreeable to put it after the imports, but before looks odd to me personally.

examples/whl.py
uranusjr: And **I** strongly disagree. Aside the readable perspective (which is kind of subjective), the div operator also presents traps on operator precedence, and encourages inefficient coding style (`foo / "bar" / "baz"`).

This is like ly not a place to have this discussion though. Can I propose we acknowledge the difference and leave it as-is? Also this might be totally moot anyway if we use our own path implementation as you proposed.

src/installer/layouts.py
gaborbernat: seems it's also where stdlib generally puts it https://github.com/python/cpython/blob/master/Lib/os.py#L36 and https://github.com/python/cpython/blob/master/Lib/asyncio/__init__.py

examples/whl.py
gaborbernat: Our own path implementation would only be used on python2.7; on laters would still use pathlib. Duno for me if we use it like this we might as well use the ``os.path`` 🤷 and avoid all the problems of ``pathlib`` (like not being available on python 2)

pyproject.toml
uranusjr: I would like to make all things work before worrying about dependencies, since the implementation would provide perspective how much the dependencies are useful. This is especially true with six because we are dealing with a lot of file IO, where Python 2 and 3 behave quite differently.

(I do think we’ll eventually get rid of `pathlib2` though. I thought it’s useful at first, but found it a bit awkward to use when dealing with zip entries.)

brettcannon:
> Script generation.

What does this entail? Is this [`console_script` entry points](https://packaging.python.org/specifications/entry-points/)?

> Maybe we should put [PEP 610] somewhere instead, e.g. packaging or importlib-metadata?

Either of those sounds reasonable to me.

uranusjr:
> What does this entail? Is this [`console_script` entry points](https://packaging.python.org/specifications/entry-points/)?

Yup.

examples/whl.py
uranusjr: I tried to trace this, and… it takes me straight back to the initial implementation of wheel support! cc @dholth

https://github.com/pypa/pip/commit/aa5b33dbf4907920c3d6cec42f1385b08096ed07#diff-2695f32c4432acd141c3dbe7e7e3a6b0R152

examples/whl.py
uranusjr: I still want something similar to `PosixPurePath` for `RECORD` even if we use `os.path` in most places though. Both `RECORD` and zipfile require forward slashes, and using string paths to interact with them is a recipe for bugs on Windows.

If we make our own `Path` I’d use it everywhere instead of on Python 2. We’ll be returning this type for `RECORD`, and it’d be woefully confusing if the two implementations differ too much.

Maybe we can use `zipp.Path` for this? It’s meant for zipfile, obviously, but the `RECORD` use case is *extremely* similar.

examples/whl.py
uranusjr: I do like to spell out an iterator to distinguish it from sequences. The call site needs to be aware of their different lifetime characteristics, otherwise it’s too easy for a maintainer to accidentally iterate through an iterator twice.

examples/whl.py
xavfernandez: It's likely to prevent concurrency issues, the rename operation being atomic.

gaborbernat: If so I consider this a bad approach, it silently swallows an important error. Sure you will not throw a failure... but instead, you're getting an environment that might not be what you expect. I'd rather throw the failure to notify the user. E.g. imagine two processes install in parallel different versions; with this approach, we'll not throw an error, but the created version will be a combination of version A and B. I'd rather throw and let the user know that something bad is happening.

gaborbernat: Not to mention that while you might not get failure on the ``.dist-info`` file, you can still get the same concurrency failure on ``.dist-info.tmp`` file.

uranusjr:
I’m going to separate this and submit them as individual PRs instead. This contains at least two entirely different concerns.

PR #4, opened 2020-05-02
Basic script generation

uranusjr:
Launcher executables are vendored from [simple_launcher](https://bitbucket.org/vinay.sajip/simple_launcher/); there’s a nox job to pull down new versions when they are available.

I opted to not handle entry point parsing because it’s already implemented by `importlib.metadata` and `pkg_resources`, but I let `Script` take plain strings so both can interface with it.


pyproject.toml
```diff
@@ -11,7 +11,8 @@ description-file = "README.md"
 classifiers = ["License :: OSI Approved :: MIT License"]
 requires-python = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
 requires = [
-  "pathlib2; python_version < '3.4'"
+  "importlib-resources; python_version < '3.7'",
```
uranusjr, line 14: Sorry @gaborbernat I’m adding even more dependencies 😛

uranusjr: I thought some more, but this one is pretty difficult. There are currently three ways to do this: `importlib.resources`, `pkg_resources`, and `pkgutil`, all with different advantages and downsides. Maybe it would be best if the user can tell the installer what to use, but I’m not quite sure if the complexity is worth it. This is yet another thing to be considered after features are in place.

gaborbernat: I'm happy with ``importlib-resources`` for now 👍

uranusjr:
I vaguely remember @zooba mentioned somewhere we can add a manifest to the executable. Is this something we can/should support?

src/installer/scripts.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if launcher is None:
+            return (self.name, shebang + b"\n" + code)
+
+        stream = io.BytesIO()
+        with zipfile.ZipFile(stream, "w") as zf:
+            zf.writestr("__main__.py", code)
+        name = "{}.exe".format(self.name)
+        data = launcher + shebang + b"\n" + stream.getvalue()
+        return (name, data)
```
uranusjr, line 116: I wonder if it’s worthwhile to return an iterator instead, so we can use `itertools.chain()` and avoid building the binary data in memory. In most scenarios people will be writing this directly to file anyway.

uranusjr, line 116: I wonder if there’s a way to stream the data instead to avoid building bytes in memory. In most scenarios people are going to write the return value directly to a file anyway.

pyproject.toml
brettcannon: I will say I'm working towards deprecating `pkgutil` someday.

src/installer/scripts.py
```diff
@@ -0,0 +1,116 @@
...
+    executable_bytes = six.ensure_binary(executable, encoding="utf-8")
+    if kind:  # The launcher can just use the command as-is.
+        return b"#!" + executable_bytes
+    if _is_executable_simple(executable_bytes):
+        return b"#!" + executable_bytes
+
+    quote = getattr(shlex, "quote", _quote_compat)
+    quoted = six.ensure_binary(quote(executable), encoding="utf-8")
+    # I don't understand a lick what this is trying to do.
+    return b"#!/bin/sh\n'''exec' " + quoted + b' "$0" "$@"\n' + b"' '''"
```
FFY00, line 76: Can you explain how the script is supposed to work? This seems a bit weird, the script will call the executable passing the script name as the first argument. So:

```sh
/path/to/my/executable script.sh arg1 arg2 arg3 ...
```

Also, this does not build the shebang, it builds the whole script. And I am not even sure why it is needed in the first place...

brettcannon:
> Would it be possible to open an issue to discuss how this is supposed to work and the overall architecture?

Feel free to open an issue if you want to start a discussion about the architecture, but this is definitely a "the person doing the work is driving the design" situation. 😉 

> I don't think blatantly copying things from distlib is a good idea.

But neither is reinventing the wheel (ha-ha, pun 😉 ). A lot of this stuff is old and has odd subtleties that people have relied on for literally over a decade, so writing quirky code from scratch  is not always the best option if you want to avoid breaking folks.

FFY00:
> Feel free to open an issue if you want to start a discussion about the architecture, but this is definitely a "the person doing the work is driving the design" situation. wink

Sure, I just asked @uranusjr to do it as he was the one implementing this.

> But neither is reinventing the wheel (ha-ha, pun wink ). A lot of this stuff is old and has odd subtleties that people have relied on for literally over a decade, so writing quirky code from scratch is not always the best option if you want to avoid breaking folks.

My point here was mainly, lets try to understand the code before merging it :smile:. For eg. the shell script I commented about is not doing anything, it can be replaced with a shebang.

uranusjr:
My intention was to use this as the issue, since it is usually easier to talk about this kind of things when we have something to start with. In fact, your comment fills in the exact missing piece I had, and now I can confidently say I understand the whole implementation!

Script generation is, in essence, taking a set of definitions, and create an executable file for the users to run. The implementation offers three kinds of executables:

For Windows, you need an exe *launcher*. There are 6 variants depending on whether the script is console- or GUI-based (corresponding to Setuptools’s `console_scripts` and `gui_scripts` sections), cross three architectures (Intel x86, AMD64, and ARM). These are generated by appending a shebang and a zip containing the script to a [pre-built executable](https://bitbucket.org/vinay.sajip/simple_launcher).

For other platforms, you need an executable script file. There are two variants to this. The default is simple enough—write a shebang, newline, and the script you want to run. There are situations, however, that this won’t work. Different operating systems treat space in shebang differently (none allows quotes), and have trouble when the path to the executable is too long. So distlib does something clever in those cases. Say you want an entry point foo to run the following code on `/path to/python`:

```python
import sys
print(sys.argv)
```

This is what distlib generates:

```sh
#!/bin/sh
'''exec' '/path to/python' "$0" "$@"
' '''
import sys
print(sys.argv)
```

This can be simplified as below:

```sh
exec '/path to/python' "$0" "$@"  # The first two single quotes cancel out.
# Everything below is irrelavent since `exec` ends this process.
```

Which means you’ll be executing the exact same file, this time with Python:

```python
#!/bin/sh
'''exec' '/path to/python' "$0" "$@"
' '''
import sys
print(sys.argv)
```

The first two lines are interpreted as a multi-line string in Python, and you get the correct result.

We definitely should document this somewhere.

FFY00:
Why is scripts generation needed? Wheel files are uncompressed zip, my (naive?) approach would be to get a list of files, check for conflicts in the filesystem, act (continue/fail) according to the arguments passed, and copy the files to the filesystem. Am I missing context here?

uranusjr:
Wheels do not contain the actual scripts defined in `entry_points`, only the definition in an `entry_points.txt` as metadata. The actual runnables are generated on installation. This is how you get the command `jupyter` when you `pip install jupyter`, for example.

FFY00:
Okay, that makes sense :smile:. But the script generation for Linux seems a bot ridiculous. Any reason we are not running the file as python directly?

```
#!/usr/bin/env python

import sys
print(sys.argv)
```

Or better, since we have the python path:

```
#!/path to/python

import sys
print(sys.argv)
```

gaborbernat:
@FFY00 there's some complication that you need to load the correct module and run the correct function, and then you need to handle shebang length limitations too (210 lines); so generating a script can get complicated. And obviously there are virtual environments that are not on the path, so ``#!/usr/bin/env python`` won't do.

uranusjr:
Your `#!/path to/python` example also won’t work on many Unix operating systems, including (IIRC) macOS. It will make them invoke an executable `/path` with argument `to/python`. m

FFY00:
You are correct. This seems to be the moist reasonable solution, as hacky as it is.

zooba:
> I vaguely remember @zooba mentioned somewhere we can add a manifest to the executable. Is this something we can/should support?

The executables in this PR already have manifests. I don't think anything else is particularly required. (I haven't checked to see whether they use relative paths or if you have to embed full paths into the file, but hopefully they're relative.)

src/installer/scripts.py
gaborbernat: Shebang has a path length limit.

eli-schwartz:
@FFY00,

In the `_is_executable_simple()` function this checks to see if the desired python interpreter is suitable for use as a shebang directly, and if so uses that as-is.

The sh hacks are only used if you do something crazy like have your python executable in a path with spaces or in a really long, subnested path, so essentially the same thing pip does, and better than what setuptools does (which is to blindly install a shebang with `#!"/path/to/python shell with spaces"`, yes, you read those quote chars correctly... of course, no one uses such crazy paths, right?).

It's actually the portable way to write scripts if you know the path is completely insane, so it seems fine to me.

For the curious, various limitations to a shebang are discussed on this page: https://www.in-ulm.de/~mascheck/various/shebang/

Including a table describing max length of shebangs (127 seems reasonable for any OS one is likely to run today).

FFY00:
Well, you could do this:

```sh
#!/bin/sh
'path to/my/python' -c "
def foo():
    return 'bar'
print(foo())
"
```

But then you would need to make sure you escape the quotes.

eli-schwartz:
Quote escaping, the general ugliness of the process tree if you inline a script into -c, and the fact that this doesn't scale well... best to avoid it IMO. :p

One could do so, but there's no inherent advantages to it.

pradyunsg:
I think the code in this PR looks fine and is "battle tested" well enough. :)

src/installer/scripts.py
uranusjr: It still makes sense to do this IMO, but the API will be different (because `importlib.resources.open_binary()` is a context manager). So the two functions can co-exist, one (in this PR) returning `bytes` and the other a stream.

With that conclusion, I think all topics are covered for script generation, and this is ready to go in.

src/installer/scripts.py
```diff
@@ -0,0 +1,120 @@
...
+        self.module = module
+        self.attr = attr
+
+    def __repr__(self):
+        # type: () -> str
+        return "Script(name={!r}, module={!r}, attr={!r}".format(
+            self.name, self.module, self.attr,
+        )
+
+    def build(self, executable, kind):
```
uranusjr, line 98: Thinking about this a bit more, it feels wrong to include the gui/console dimension in ``kind``; the distinction is made in entry scripts sections, so the flag really should be set on the `Script` class instead. I want to make some modifications:

1. Make `Script.__init__()` accept an additional parameter to distinguish between a script in  `gui_scripts` and `console_scripts`.
2. Reduce the `kind` parameter here to only `""`, `"32"`, `"64"`, and `"arm"`. The launcher name is generated based on this kind parameter and the gui/console flag on the Script instance.

But @pradyunsg since you’ve already approved this PR, I’m fine to make the modification after this goes into master as well.

pradyunsg: Go ahead and make this change. I'm in no hurries to merge. :)

uranusjr:
I’m happy with this now!

PR #6, opened 2020-05-04
Implement our own lightweight Path

uranusjr:
Discussion at https://github.com/pradyunsg/installer/pull/3/files#r418961489

### DIFF
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -1,15 +1,15 @@
 import csv
 import warnings
 
-from installer._compat import pathlib
 from installer._compat.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Iterator, Optional
+    from typing import Iterable, Iterator, Optional
 
 
 __all__ = [
     "Hash",
+    "Path",
     "RecordItem",
     "SuperfulousRecordColumnsWarning",
     "parse_record_file",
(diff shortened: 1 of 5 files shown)


src/installer/records.py
```diff
@@ -8,17 +9,53 @@
...
+
+    def __init__(self, *parts):
+        # type: (*str) -> None
+        self._parts = tuple(self._iter_parts(parts))
+
+    @staticmethod
+    def _iter_parts(parts):
+        # type: (Iterable[str]) -> Iterator[str]
+        for part in parts:
+            for p in part.split("/"):
```
gaborbernat, line 38: still, not a fan of a single char variable if it's not inside a comprehension; and also find the iter keyword in there useful, especially with type annotations.

uranusjr: Not even for a variable that’s only used in the immediately two lines after? IMHO this is a bit obessesive, but I’m OK if you feel strongly. What would be a good alternative? I can think of something like `split_part`, but that doesn’t read more easily to me.

Speaking of obssesive, I do feel very strongly about the `iter` part. The type hint is not useful on the call site, where the information is important. And it is very possible to write code that escape type checks but entirely not make sense:

```python
parts = self._split_parts(parts)  # This returns an iterator.
for part in parts:
   # ...
for part in parts:  # This is wrong.
   # ...
```

Sticking `iter` in the name is the best way IMO to avoid this class of coding errors.

uranusjr:
I just realised it is not suitable to use `str` as the type for each path component. `RECORD` entries are encoded in UTF-8 (not actually specified anywhere, but implemented by `importlib.metadata`), and so are file entries in the wheel (as specified in PEP 427). But installed path needs to be calculated using platform filesystem encoding (obviously). So the suitble type here should be `unicode` on Python 2 instead, i.e. we should use `six.text_type`.

pradyunsg:
With #11 merged, I don't think we need to be doing this.

PR #9, opened 2020-05-28
records: add RecordItem.validate()

FFY00:
I opted for `validate` to receive `bytes` as this should be compatible with both `ZipFile` and normal files, but this *could* be a problem with big files. What do you think?

If we choose to receive a `BinaryIO` instead of `bytes` it could be tricky to get the size. I guess if the object is seekable, we can seek to the end and use `tell`.

### DIFF
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -1,7 +1,9 @@
 """Utilities for parsing and handling PEP 376 RECORD files.
 """
 
+import base64
 import csv
+import hashlib
 import os
 
 from installer._compat.typing import TYPE_CHECKING
@@ -64,6 +66,18 @@ class Record(object):
         return "Record(path={!r}, hash_={!r}, size={!r})".format(
             self.path, self.hash_, self.size,
         )
+
+    def validate(self, data):
+        # type: (bytes) -> bool
+        if self.size is not None and len(data) != self.size:
+            return False
+
+        if self.hash_:
+            digest = hashlib.new(self.hash_.name, data).digest()
+            value = base64.urlsafe_b64encode(digest).decode("ascii").rstrip("=")
+            return self.hash_.value == value
+
+        return True
 
     @classmethod
     def from_elements(cls, path, hash_, size):
--- a/tests/test_records.py
+++ b/tests/test_records.py
@@ -66,6 +66,44 @@ class TestRecord:
     def test_valid_elements(self, path, hash_, size):
         Record.from_elements(path, hash_, size)
 
+    @pytest.mark.parametrize(
+        ("elements", "data", "expected"),
+        [
+            (
+                ("test1.py", "sha256=Y0sCextp4SQtQNU-MSs7SsdxD1W-gfKJtUlEbvZ3i-4", 6),
+                b"test1\n",
+                True,
+            ),
+            (
+                ("test2.py", "sha256=fW_Xd08Nh2JNptzxbQ09EEwxkedx--LznIau1LK_Gg8", 6),
+                b"test2\n",
+                True,
+            ),
+            (
+                ("test3.py", "sha256=qwPDTx7OCCEf4qgDn9ZCQZmz9de1X_E7ETSzZHdsRcU", 6),
+                b"test3\n",
+                True,
+            ),
+            (
+                ("test4.py", "sha256=Y0sCextp4SQtQNU-MSs7SsdxD1W-gfKJtUlEbvZ3i-4", 7),
+                b"test1\n",
+                False,
+            ),
+            (
+                (
+                    "test5.py",
+                    "sha256=Y0sCextp4SQtQNU-MSs7SsdxD1W-gfKJtUlEbvZ3i-4",
+                    None,
+                ),
+                b"test1\n",
+                True,
+            ),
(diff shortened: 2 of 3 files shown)


FFY00:
About the falake8 failing, should I stick `# noqa: E501` everywhere or do you wan to handle this in other way?

src/installer/records.py
```diff
@@ -49,6 +61,17 @@ def __repr__(self):
...
         )
 
+    def validate(self, data):
+        # type: (bytes) -> bool
+        if self.size is not None and len(data) != self.size:
+            return False
+
+        if self.hash_:
+            hash = hashlib.new(self.hash_.name, data)
+            return hash.digest() == self.hash_.digest
```
uranusjr, line 71: I personally would favour just compare the base64-encoded values, i.e. instead of decoding the `Hash.value` back, encode the incoming has instead:

```python
digest = hashlib.new(self.hash_.name, data).digest()
value = base64.urlsafe_b64encode(digest).decode("ascii").rstrip("=")
return self.hash_.value == value
```

The difference is minimal, but I like to stick to the security best practice of always comparing the encoded result.

tests/test_records.py
```diff
@@ -81,6 +86,92 @@ def test_parse_wheel_record_drop_superfulous():
...
+        ),
+        (
+            "sha256=sGatYQZagyus0JT42xtMA0xCrBK0lwjhWTSALcS2atA",
+            b"\xb0f\xada\x06Z\x83+\xac\xd0\x94\xf8\xdb\x1bL\x03LB\xac\x12\xb4\x97\x08\xe1Y4\x80-\xc4\xb6j\xd0",
+        ),
+        (
+            "sha256=yralTN5PtS-rXno4zFzQj6oNbfpK7PK4rzD8gVUg95Y",
+            b"\xca\xb6\xa5L\xdeO\xb5/\xab^z8\xcc\\\xd0\x8f\xaa\rm\xfaJ\xec\xf2\xb8\xaf0\xfc\x81U \xf7\x96",
+        ),
+    ],
```
uranusjr, line 124: Why do we need this many test cases?

uranusjr:
> About flake8 failing, should I stick `# noqa: E501` everywhere or do you want to handle this in other way?

Huh, did Black change the default line length? Because the 88 configuration in flake8 was specifically to be compatible with Black’s output. Could you try adding `line-length = 88` to Black’s configuration and see if that works?

FFY00:
I added
```
[tool.black]
line-length = 88
```
to `pyproject.toml` but it's still the same.

FFY00:
Nevermind, I did something wrong :)

FFY00:
@pradyunsg can you have a look?

pyproject.toml
```diff
@@ -22,3 +22,6 @@ test = [
   "pytest-xdist",
 ]
 doc = ["sphinx"]
+
+[tool.black]
+line-length = 88
```
pradyunsg, line 27: Please don't do this. If this is for the hashes in the decorator being too long, I'd prefer we use `# fmt: off` and `# fmt: on` comments to avoid black's strictness around how the values should be formatted.

tests/test_records.py
```diff
@@ -1,6 +1,10 @@
 import pytest
 
-from installer.records import SuperfluousRecordColumnsWarning, parse_record_file
+from installer.records import (
+    RecordItem,
```
pradyunsg, line 4: Looks like we should rebase this PR on #12 (or wait for that to merge, then rebase on master).

This would also mean changing the tests to be a method on the relevant class.

tests/test_records.py
```diff
@@ -81,6 +85,51 @@ def test_parse_wheel_record_drop_superfulous():
     assert r1.size is None
 
 
+@pytest.mark.parametrize(
+    ("record", "data", "expected"),
```
pradyunsg, line 89: Instead of constructing the `Record` in the decorator's input, can we move `Record.from_elements` into the function call? The argument/values can stay a tuple called `elements`.

This avoids assuming "`Record.from_elements` will work" for being able to run this file (which tests that `Record` class).

pradyunsg:
> Huh, did Black change the default line length?

No, black doesn't break strings when they go over the 88 character limit.

pyproject.toml
pradyunsg: Or, we can separately disable E501 in flake8, since formatting code via black basically makes that error redundant.

FFY00: Disabling `E501` seems good to me.

FFY00:
Rebased on top of #12.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+ignore = E501
```
pradyunsg, line 2: Please do this in the .flake8 file instead.

FFY00: Ah sorry, `ls` doesn't show `.*` so I missed it :smile:.

FFY00:
Rebased on top of master.

pradyunsg:
@FFY00 I just realized, we should probably move the logic within `if self.hash_` into a `Hash.validate` method and call that.

FFY00:
That is fine by me. I will open a PR later.

PR #10, opened 2020-05-28
Cache nox & pip in CI

FFY00:
From `ci: cache nox`
> Right now, we recreate the nox environment on every CI run, which is wasteful
> as it can take a while to to create it.
>
> This patch caches the nox environment so that it doesn't need to be
> recreated on every CI run. The cache will only be updated once a day.
>
> If at some point we save a bad environment or the cached environment
> becomes incompatible with newer nox versions, we should bump the
> version (currently v1) in the cache key, to force the cache to be
> recreated.
>
> Signed-off-by: Filipe Laíns <lains@archlinux.org>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,16 +15,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    # Get nox cache
-    - name: Save date (for cache)
-      id: date
-      run: echo "::set-output name=date::$(date +%F)"
-    - uses: actions/cache@v1
-      with:
-        path: ${{ github.workspace }}/.nox
-        key: nox-v1-${{ runner.os }}-${{ steps.date.outputs.date }}
-        restore-keys: nox-v1-${{ runner.os }}
-
     # Get all the Python versions
     - uses: actions/setup-python@v1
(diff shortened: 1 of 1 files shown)


pradyunsg:
Not waiting on additional reviews since:

- This is simple enough.
- If this hurts more than it helps, we can iterate on it.
- We always have the option of reverting, if we really need to.

pradyunsg:
Looks like nox isn't happy about being cached.

FFY00:
Are you sure it's because of nox? What is failing here is `pip install .[test]`, I am not sure why that would fail.

Also, I think nox/tox is really helpful to run tests locally but I don't think it makes that much sense for a CI.
My proposal would be something like this: https://github.com/FFY00/python-install/blob/master/.github/workflows/checks.yml
This is the result: https://github.com/FFY00/python-install/actions/runs/125454437
As you can see, everything runs in parallel, it's much faster (the CI setup routines are optimized for fast setup) and results in a more readable CI output. I just thought it was worth asking this before starting looking into why nox is failing.

pradyunsg:
It's since `.nox` shouldn't have been cached. #15 reverted that part of this change.

PR #11, opened 2020-06-07
Drop pathlib2 dependency and merge `installer._compat.{builtins, typing}`

pradyunsg:
Follow up on #2 

This also adds `Text` and `Binary` type names, which are better IMO. (and follow mypy's `Text` type)


uranusjr:
Quoting myself from https://github.com/pradyunsg/installer/pull/6#issuecomment-625604376:

> It is not suitable to use `str` as the type for each path component. `RECORD` entries are encoded in UTF-8 (not actually specified anywhere, but implemented by `importlib.metadata`), and so are file entries in the wheel (as specified in PEP 427). But installed path needs to be calculated using platform filesystem encoding (obviously). So the suitble type here should be `unicode` on Python 2 instead.

In term of this pull request, `FSPath` should always be `Text`, never `Binary`.

pradyunsg:
Can we say that paths in all wheel files must be UTF-8 encoded?

If so, then it makes sense to drop Binary from the declarations. It'll also simplify the handling of the INSTALLER file in the implementation, since I can rely on the fact that it'll be `u"INSTALLER"`.

uranusjr:
There was a discussion on this. I can’t recall where it’s made, but basically

1. Neither the wheel specification nor PEP 376 specifies what encoding files in the dist-info directory should be.
2. All existing metadata introspection implementations (e.g. `pkg_resources` and `importlib.metadata`) uses UTF-8.
3. All existing wheel creators use UTF-8 to write them
4. pip historically uses platform encoding to install wheels (this improved recently with my series of wheel encoding fixes). The combination of 2. and this basically means all pip-installable dist-info files are only ASCII-compatible.

So I’d say it should be expected at this point metadata files in a public wheel file should be in UTF-8, and a generic tool should assume this at all times.

pradyunsg:
Sounds fair to me. ^.^

src/installer/_compat/typing.py
```diff
@@ -1,6 +1,37 @@
...
     TYPE_CHECKING = False
 
-__all__ = ["TYPE_CHECKING"]
+if TYPE_CHECKING:
+    from typing import Union
+
+    if sys.version_info[:2] >= (3, 4):
+        from pathlib import Path
+
+        FSPath = Union["Text", "Binary", Path]
```
pradyunsg, line 28:
```suggestion
        FSPath = Union["Text", Path]
```

src/installer/_compat/typing.py
```diff
@@ -1,6 +1,37 @@
...
+    else:
+        FSPath = Union["Text", "Binary"]
```
pradyunsg, line 30:
```suggestion
        FSPath = "Text"
```

pradyunsg:
Okai, I'm gonna go ahead and merge this, since no one has stated anything negative about this plan of action / PR. Plus, rebasing this and #12 to work around the CI issues introduced via #10 has been a pain. :)

PR #12, opened 2020-06-07
Rework installer.records

pradyunsg:
Follow up on #2, builds upon #11

### DIFF
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -1,3 +1,6 @@
+"""Utilities for parsing and handling PEP 376 RECORD files.
+"""
+
 import csv
 import os
 
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -64,23 +64,47 @@ class Record(object):
 
     @classmethod
     def from_elements(cls, path, hash_, size):
-        # type: (str, str, str) -> Record
-        """Build a Record from parsing elements in a record row.
+        # type: (FSPath, str, str) -> Record
+        """Build a Record object, from values of the elements.
+
+        All arguments must be in string form, except `path` which can also be a
+        ``pathlib.Path`` instance on Python 3.
 
         Typical usage::
 
             reader = csv.reader(f)
             for row in reader:
-                record = Record.parse(row[0], row[1], row[2])
+                record = Record.from_elements(row[0], row[1], row[2])
 
-        All arguments are in string form. Meaning of elements are specified in
-        PEP 376. Raises ``ValueError`` if any of the elements is invalid.
+        All arguments are in string form. Meaning of each element is specified in
+        PEP 376. Raises ``InvalidRecord`` if any element is invalid.
         """
-        return cls(
-            path=path,
-            hash_=Hash.parse(hash_) if hash_ else None,
-            size=int(size) if size else None,
-        )
+        # Validate the passed values.
+        issues = []
+
+        if not path:
+            issues.append("`path` cannot be empty")
+
+        if hash_:
+            try:
+                hash_value = Hash.parse(hash_)  # type: Optional[Hash]
+            except ValueError:
+                issues.append("`hash` does not follow the required format")
+        else:
+            hash_value = None
+
+        if size:
+            try:
+                size_value = int(size)  # type: Optional[int]
+            except ValueError:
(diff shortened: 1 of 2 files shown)


src/installer/records.py
```diff
@@ -36,50 +49,74 @@ def parse(cls, h):
...
+            message = "Row Index {}: expected 3 elements, got {}".format(
+                row_index, len(elements)
             )
-        try:
-            record = RecordItem.parse(row[0], row[1], row[2])
-        except (IndexError, ValueError):
-            raise ValueError("invalid row {}: {!r}".format(row_index, row))
+            raise InvalidRecord(elements=elements, issues=[message])
+
+        record = Record.from_elements(elements[0], elements[1], elements[2])
```
FFY00, line 121: nitpick: why not yield here?

src/installer/records.py
```diff
@@ -36,50 +49,74 @@ def parse(cls, h):
...
+                size_value = int(size)  # type: Optional[int]
+            except ValueError:
+                issues.append("`size` cannot be non-integer")
+        else:
+            size_value = None
+
+        if issues:
+            raise InvalidRecord(elements=(path, hash_, size), issues=issues)
+
+        return cls(path=path, hash_=hash_value, size=size_value,)
```
uranusjr, line 106: Extra comma. When on earth is Black going to fix this bug already?

src/installer/records.py
pradyunsg: Out of habit, since it makes it easier for git to anchor properly. It doesn't matter in this case, but as I said, habits. :)

src/installer/records.py
pradyunsg:
```suggestion
        return cls(path=path, hash_=hash_value, size=size_value)
```

pradyunsg:
@FFY00 Do you mean breaking up 27b1be7?

@uranusjr Lemme know if you have any other suggestions here. ^>^

FFY00:
Yes, it could be split up into smaller changes. But you don't have to, it is fine as it is. I mainly commented for you to take this into account in the future.

pradyunsg:
> Yes, it could be split up into smaller changes.

Yea, I was just being lazy. I've gone ahead and done this now. Hopefully, I didn't bork the changes. :)

pradyunsg:
I'll hold off on merging this PR, so that hopefully a few more :eyes: can review this. :)

src/installer/records.py
```diff
@@ -36,50 +50,77 @@ def parse(cls, h):
...
+
+
+def parse_record_file(rows):
+    # type: (Iterator[str]) -> Iterator[Record]
+    """Parse a RECORD file, provided as an iterator of record lines.
+    """
+    reader = csv.reader(rows, delimiter=",", quotechar='"', lineterminator=os.linesep)
+    for row_index, elements in enumerate(reader):
+        if len(elements) != 3:
```
uranusjr, line 119: I wonder if forward-compatibility should be an consideration here, i.e. if the wheel spec add more columns to the record, the installer should be tolerant to the optional values. Hard-error on less than three columns is perfectly fine, of course.

pradyunsg: Maybe? I think it's better to be strict on what we have today, instead of trying to anticipate how we might change things in the future and accommodating for them now.

I'm pretty sure installer will need to be updated for some reason or another, when a new wheel format version is created. #16 has a Wheel-Version check embedded, so if we don't change that, then there's a clear definition of which versions of the wheel format are supported in a certain version of this library.

Support for newer Wheel format versions needs a newer versions of this library - that's a very reasonable approach to compatibility with newer wheel versions IMO.

uranusjr: OK then.

pradyunsg: @uranusjr That might be since you don't have the commit/triage bit on this?

pradyunsg:
Alrighty, I'm gonna go ahead and merge this, since no one has raised any new concerns, and this will make the #9 and #16 diffs cleaner.

pradyunsg:
Thanks for the reviews @FFY00 and @uranusjr! And also for the initial implementation @uranusjr! ^>^

PR #14, opened 2020-06-07
Naming changes in installer.scripts

pradyunsg:
Renames:

- `tools/getlaunchers.py` -> `tools/update_launchers.py`
- `Script.build` -> `Script.generate`

We should really avoid using the phrase "build" in this package. That word is used often enough that it can get confusing what's being talked about. :)


uranusjr:
I am -0 to `generate` since the verb is related to a generator.

eli-schwartz:
The word "generate" IMHO cannot be confused with the python term "generator" as the latter does not really have any semantic meaning when spelled differently in any way shape or form...

On the other hand, if the name is concerning, I'd like to point out this doesn't build or generate or anything of the sort, anywhere here. It doesn't matter how many times you do it, you won't rebuild or regenerate a "script", because the results of the previous invocation were ephemeral, and the function name, both of them, might fool the casual reader into thinking that it is backed by a file.
It *formats* the `_SCRIPT_TEMPLATE` and returns a string.

How about `Script.get_data`?

brettcannon:
Other naming options are "create" or "as_script".

FFY00:
Also, `construct`.

pradyunsg:
`create` sounds good to me.

Unless someone raises concerns with that name in the next day or so, I'm gonna swap to that.

eli-schwartz:
It still feels weird that way. Think about the API you're providing, and imagine you're a new user:

```
import sys
import installer.scripts
s = installer.scripts.Script('foo', 'foobar', 'main', 'console' )
s.build(sys.executable, kind='posix')
```
What do you expect s.build() to do? Well, build sounds like it will act in the build stage, so we want to change that, right?

Okay, let's try again:

```
s.create(sys.executable, kind='posix')
```

"Okay, I've created it, now where is it?"

I still think get_data is clearer:

```
ret = s.get_data(sys.executable, kind='posix')
```

"Okay, I've gotten the data contents of the Script, I should probably check type(ret) and figure out how I should use it; it will probably be some data blob."

Mentioning the "data" somewhere in the method name seems like a useful way to avoid potential confusion later on.

pradyunsg:
OTOH, I still think `ret = s.generate(...)` is good enough too. If no one's a -1 on that, I'll just merge this PR. :)

pradyunsg:
Thanks for the reviews and thoughts everyone! ^>^

Issue #1, opened 2020-04-11
Scope of this project and API design


pradyunsg:
This project is born out of https://discuss.python.org/t/3869/. Let me know here if you participated there and want push access.

As a first step, we should figure out what *exactly* do we want this project to be. :)

gaborbernat:
Sign me up; though for virtualenv I'd have some perhaps unusual requirements (which may end up falling out of scope, in which case virtualenv could not use it to be fair):

- Python 2.7 support,
- being modular enough to allow performance optimizations... such as creating wheel images that no longer need extraction/setup but can be reduced to just copy folder A to folder B (avoids need to regenerate pyc files, extend RECORDS, doing a copy is faster on Windows as the anti-virus already checked the sources so the new destination is less suspicious).

uranusjr:
If we’re going the sans-I/O route, it would make most sense to me to structure this project as a collection of tools—parsers and “unparsers” (is this an actual word?) needed to convert a wheel into an “installation”.

It likely also makes sense to also implement a wrapper API that calls those tools in order, which can both be used as-is, or serve as an example if someone wants to plug the tools into a different I/O source. The API @xavfernandez used in [cric](https://github.com/xavfernandez/cric) makes a lot of sense to me, although I’d prefer it taking a file-like object, and make the scheme a structured object.

I don’t have experience with sans-I/O interface design, so it makes most sense to me if we can implement the wrapper API first with the [flow suggested by @dholth](https://discuss.python.org/t/creating-a-package-to-just-install-a-wheel/3869/19?u=uranusjr), and slowly pick it apart into the two-layer design. But I’ll happily follow and learn from people who do know how this kind of interface works 🙂

EpicWink:
The README says the project is for installing Python wheels

1. Installation is an action, which like any function, has input, output, side-effects and an environment to run in. These should be well-defined
2. What extra functionality is desired? eg uninstallation, hash-checking (etc), inspection
3. Will the project contain an app (eg executable from command-line), a library (to be used by other libraries and apps), or both? If both, will the app and library be the same code?
4. Will the installer have configuration, like pip, or be "this is how we do things and you will like it", like [black](https://black.readthedocs.io/en/stable/). If so, is this configuration designed to be interfaced with by other programs (command-line, config-file, or otherwise), or just squishies (humans)?
5. Will this support any network requests at all, will it even use network to download a requested package to install, or will it only act on local data?

> “unparsers”

"compiler", or "formatter", depending on context

[...]

PR #16, opened 2020-06-17
Initial implementation

pradyunsg:
Built on top of #12 

Filing early to get inputs on this design / approach. As of right now, the design has 4 interacting components: the `Installer` class encapsulates all the installation logic, interacts with all the other abstractions and is the main entry point. `WheelSource` and `Destination`, which do exactly what their name suggests. And, finally, there's the coolest kids on the block, `Validator`s.

`WheelSource` represents a wheel file (like a .whl file on disk). It provides random access to dist-info files, and sequential access to *all* the files within the wheel. This lets `Installer` read the metadata files it needs (only WHEEL), any potential validators to read other metadata (like RECORD files), while making it straightforward to consume the contents of the wheel.

`Destination` handles all the file-writing stuff (so all the I/O) and has the job of rewriting the RECORD file appropriately.

`Validator`s are... `Callable[[WheelSource], None]` which can raise an error to signal not-valid. That's it.

---

One of the quirks of this approach/design, is that we don't actually "unpack" the wheel root into the corresponding scheme, but instead figure out the correct place to put the files the moment we see them. This lets us simplify the Destination API, and avoids an extra step (potentially reducing I/O operations).

---

Two things that I am wondering:

- Maybe we should have some way to preserve `stat(...).st_mtime` from `WheelSource` to `Destination`?
- Is there a better abstraction to pass around than BufferedReader?

### DIFF
--- a/src/installer/_core.py
+++ b/src/installer/_core.py
@@ -1,0 +1,113 @@
+"""Wheel Installation logic, sans I/O
+"""
+
+import posixpath
+
+from installer._compat import StringIO
+from installer._compat.typing import TYPE_CHECKING
+from installer.parsers import parse_metadata_file
+from installer.protocols import Destination, Scheme, WheelSource
+from installer.validation import InvalidWheelSource
+
+if TYPE_CHECKING:
+    from typing import Callable, List, Optional
+
+    from installer._compat.typing import Text
+
+    Validator = Callable[[WheelSource], None]
+
+
+__all__ = ["Installer"]
+
+
+class Installer(object):
+    """Implements wheel installation logic
+
+    Supports Wheel version 1.0 (PEP 427).
+    """
+
+    def __init__(self, name, validators):
+        # type: (Optional[Text], List[Validator]) -> None
+        super(Installer, self).__init__()
+        self.name = name
+        self._validators = validators
+
+    def install(self, source, destination, requested=False):
+        # type: (WheelSource, Destination, bool) -> None
+        """Install a wheel, as described by ``source``, to ``destination``.
+        """
+
+        # Process the WHEEL file
+        wheel_metadata_stream = source.open_dist_info(u"WHEEL")
+        metadata = parse_metadata_file(wheel_metadata_stream)
+
+        # Ensure compatibility with this wheel version.
+        if metadata["Wheel-Version"] != "1.0":
+            raise InvalidWheelSource(
+                source, "Incompatible Wheel-Version: only support version 1.0",
+            )
+
+        self._pass_through_validators(source)
+
+        # Determine where archive root should go.
+        if metadata["Root-Is-Purelib"]:
+            root_scheme = Scheme.purelib
+        else:
+            root_scheme = Scheme.platlib
+
+        # Unpack all the files, directly into the correct scheme.
+        for path, stream in source.iter_files():
+            scheme = self._determine_scheme(path, source=source, fallback=root_scheme)
+            destination.write_file(scheme=scheme, path=path, stream=stream)
+
+        # Write INSTALLER, if name is provided.
+        if self.name is not None:
+            destination.write_file(
+                scheme=root_scheme,
+                path=posixpath.join(source.dist_info, u"INSTALLER"),
+                stream=StringIO(self.name),
+            )
+        # Write REQUESTED, if requested.
+        if requested:
+            destination.write_file(
+                scheme=root_scheme,
+                path=posixpath.join(source.dist_info, u"REQUESTED"),
+                stream=StringIO(u""),
+            )
+
+        # Rewrite the RECORD at the end.
+        destination.rewrite_record(scheme=root_scheme)
+
+    def _pass_through_validators(self, source):
+        # type: (WheelSource) -> None
+        errors = []
+        for validator in self._validators:
+            try:
+                validator(source)
+            except InvalidWheelSource as e:
+                errors.append(e)
+
+        if errors:
+            raise InvalidWheelSource.from_multiple(source, errors)
+
+    def _determine_scheme(self, path, source, fallback):
+        # type: (FSPath, WheelSource, Scheme) -> Scheme
+        data_dir = source.data_dir
+
+        # If it's in not `{distribution}-{version}.data`, then it's in root_scheme.
+        if not posixpath.commonprefix([data_dir, path]) == data_dir:
+            return fallback
+
+        left, right = posixpath.split(path)
+        while left != data_dir:
+            left, right = posixpath.split(left)
+
+        scheme_name = right
+        if scheme_name not in Scheme.__members__:
+            msg_fmt = u"{path} is not contained in a valid .data subdirectory."
+            raise InvalidWheelSource(
+                source,
+                msg_fmt.format(path=path)
+            )
+
+        return Scheme.__members__[scheme_name]
--- a/src/installer/_core.py
+++ b/src/installer/_core.py
@@ -21,7 +21,7 @@ __all__ = ["Installer"]
 
 
 class Installer(object):
-    """Implements wheel installation logic
+    """Implements wheel installation logic.
 
     Supports Wheel version 1.0 (PEP 427).
     """
@@ -37,7 +37,7 @@ class Installer(object):
         """Install a wheel, as described by ``source``, to ``destination``.
         """
 
-        # Process the WHEEL file
+        # Process the WHEEL file.
         wheel_metadata_stream = source.open_dist_info(u"WHEEL")
         metadata = parse_metadata_file(wheel_metadata_stream)
 
--- a/src/installer/protocols.py
+++ b/src/installer/protocols.py
@@ -16,8 +16,8 @@ Scheme = Enum("Scheme", ["purelib", "platlib", "headers", "scripts", "data"])
 
 @with_metaclass(abc.ABCMeta)
 class WheelSource(object):
-    """Represents an installable wheel.
-    """
+    
+    """Represents an installable wheel."""
 
     def __init__(self, distribution, version):
         # type: (Text, Text) -> None
@@ -27,14 +27,12 @@ class WheelSource(object):
 
     @property
     def dist_info_dir(self):
-        """Name of the dist-info directory.
-        """
+        """Name of the dist-info directory."""
         return u"{}-{}.dist-info".format(self.distribution, self.version)
 
     @property
     def data_dir(self):
-        """Name of the data directory.
-        """
+        """Name of the data directory."""
         return u"{}-{}.data".format(self.distribution, self.version)
 
     @abc.abstractproperty
--- a/src/installer/protocols.py
+++ b/src/installer/protocols.py
@@ -1,0 +1,120 @@
+import abc
+from enum import Enum
+
+from installer._compat import with_metaclass
+from installer._compat.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from io import BufferedReader
+    from typing import List, Sequence, Tuple
+
+    from installer._compat.typing import Text, FSPath
+
+
+Scheme = Enum("Scheme", ["purelib", "platlib", "headers", "scripts", "data"])
+
+
+@with_metaclass(abc.ABCMeta)
+class WheelSource(object):
+    """Represents an installable wheel.
+    """
+
+    def __init__(self, distribution, version):
+        # type: (Text, Text) -> None
+        super(WheelSource, self).__init__()
+        self.distribution = distribution
+        self.version = version
+
+    @property
+    def dist_info_dir(self):
+        """Name of the dist-info directory.
+        """
+        return u"{}-{}.dist-info".format(self.distribution, self.version)
+
+    @property
+    def data_dir(self):
+        """Name of the data directory.
+        """
+        return u"{}-{}.data".format(self.distribution, self.version)
+
+    @abc.abstractproperty
+    def dist_info_filenames(self):
+        # type: () -> List[FSPath]
+        """Get names of all files in the dist-info directory.
+
+        Sample usage/behavior::
+
+            >>> wheel_source.dist_info_filenames
+            ["INSTALLER", "METADATA", "WHEEL"]
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def open_dist_info(self, filename):
+        # type: (FSPath) -> BufferedReader
+        """Get readable stream of data, from ``filename`` in the dist-info directory.
+
+        Sample usage/behavior::
+
+            >>> f = wheel_source.open_dist_info("INSTALLER")
+            >>> f.read()
+            'pip'
+        """
+        raise NotImplementedError()
+
+    # All files in the wheel
+    @abc.abstractmethod
+    def iter_files(self):
+        # type: () -> Iterator[Tuple[FSPath, BufferedReader]]
+        """Sequential access to all contents of the wheel (including dist-info files).
+
+        All paths must be posix-style paths (i.e. can be used with ``posixpath`` module)
+        relative to the root of the wheel.
+
+        Sample usage/behavior::
+
+            >>> iterable = wheel_source.iter_files()
+            >>> next(iterable)
+            ('pkg-1.0.0.dist-info/METADATA', <...>)
+
+        This method may be called multiple times, and each iterator returned must
+        provide the same content upon reading from the streams.
+        """
+        raise NotImplementedError()
+
+
+@with_metaclass(abc.ABCMeta)
+class Destination(object):
+    """Represents the location for installation.
+
+    Subclasses are expected to handle script generation and rewriting of the
+    RECORD file after installation.
+    """
+
+    @abc.abstractmethod
+    def write_file(self, scheme, path, stream):
+        # type: (Scheme, FSPath, BufferedReader) -> None
+        """TODO: write a good one line description of this function.
+
+        Example usage/behavior::
+
+            >>> stream = open("__init__.py")
+            >>> dest.write_file(Scheme.purelib, "__init__.py", stream)
+
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def rewrite_record(self, scheme):
+        # type: (Scheme) -> None
+        """Rewrite RECORD, using information already provided via ``write_file`` calls.
+
+        Example usage/behavior::
+
+            >>> dest.write_file(Scheme.purelib, "__init__.py", stream1)
+            >>> dest.write_file(Scheme.purelib, "awesome_wrapper.py", stream2)
+            >>> dest.write_file(Scheme.purelib, "_awesome.pyd", stream3)
+            >>> dest.rewrite_record(Scheme.purelib)
+
+        """
+        raise NotImplementedError()
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -37,8 +37,8 @@ def parse_metadata_file(contents):
 # RECORD file parsing
 #
 class InvalidRecord(Exception):
-    """Raised when a Record is not valid, due to improper element values or count.
-    """
+
+    """Raised when a Record is not valid, due to improper element values or count."""
 
     def __init__(self, elements, issues):
         super(InvalidRecord, self).__init__(", ".join(issues))
@@ -145,8 +145,7 @@ class Record(object):
 
 def parse_record_file(rows):
     # type: (Iterator[str]) -> Iterator[Record]
-    """Parse a RECORD file, provided as an iterator of record lines.
-    """
+    """Parse a RECORD file, provided as an iterator of record lines."""
     reader = csv.reader(rows, delimiter=",", quotechar='"', lineterminator=os.linesep)
     for row_index, elements in enumerate(reader):
         if len(elements) != 3:
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -5,22 +5,37 @@ import base64
 import csv
 import hashlib
 import os
+from email.parser import FeedParser
 
 from installer._compat.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from email.message import Message
     from typing import Iterator, Optional
     from installer._compat.typing import FSPath
 
-
 __all__ = [
-    "Hash",
-    "Record",
-    "InvalidRecord",
-    "parse_record_file",
+    "parse_metadata_file", "parse_record_file", "Hash", "Record", "InvalidRecord",
 ]
 
 
+#
+# METADATA (and WHEEL) file parsing
+#
+def parse_metadata_file(contents):
+    # type: (Text) -> Message
+    """Parse a METADATA / WHEEL / PKG-INFO file.
+
+    :param contents: The entire contents of the file.
+    """
+    feed_parser = FeedParser()
+    feed_parser.feed(contents)
+    return feed_parser.close()
+
+
+#
+# RECORD file parsing
+#
 class InvalidRecord(Exception):
     """Raised when a Record is not valid, due to improper element values or count.
     """
--- a/src/installer/validation.py
+++ b/src/installer/validation.py
@@ -1,0 +1,26 @@
+"""Validation checks for wheel files.
+"""
+
+
+class InvalidWheelSource(Exception):
+    """Raised when a ``WheelSource`` is not valid.
+    """
+
+    def __init__(self, source, reason):
+        super(InvalidWheelSource, self).__init__(source, reason)
+        self.source = source
+        self.reason = reason
+
+    @classmethod
+    def from_multiple(cls, source, exceptions):
+        # type: (WheelSource, Iterable[InvalidWheelSource])
+        reason = ", ".join(e.reason for e in exceptions)
+        return cls(source=source, reason=reason)
+
+
+#
+# Basic Validators
+#
+def validate_contents_match_RECORD_file(source):
+    # TODO: implement this, after #9 is merged.
+    pass


pradyunsg:
The most important files are likely `protocols.py` (open to naming suggestions!) and `_core.py` (I like this name, let's keep this as is).

There's a few commits from #12, which don't really add much noise to the overall PR due to the `installer.{records -> parsers}` renaming.

pradyunsg:
Oh, and this code doesn't actually run, since there's missing imports and shitz. I'll polish all of that up, once it's clear whether this is an appropriate direction to invest effort into. :)

FFY00:
Overall looks good, one thing jumps out to me, why does the API user need to implement `Destination.rewrite_record`?

FFY00:
Actually, I am not entirely sure how it is supposed to work. Does it check all files and fixes up `RECORD`?

pradyunsg:
That's because only the `Destination` object has information about the file paths needed for performing the rewrite of the RECORD file. I'm 100% open to figuring out a helper to simplify end-user implementations in the common case, but within the abstractions, I don't think the responsibility for rewriting the RECORD file can fit anywhere else.

pradyunsg:
> Does it check all files and fixes up `RECORD`?

See point 4 in "spread" on PEP 427's description of how to install a wheel: https://www.python.org/dev/peps/pep-0427/#installing-a-wheel-distribution-1-0-py32-none-any-whl

Basically, the path component of each record needs to be changed. This is worth mentioning in the `rewrite_record`'s docstring.

FFY00:
> That's because only the `Destination` object has information about the "final" file paths

Then we can ask the user to define a `get_full_path` method, which could then be used by our `rewrite_record`. As an API user, I don't want to care about `RECORD` rewriting logic. This is also one of those things that people can easily get wrong, we want to prevent that.

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+__all__ = ["Installer"]
+
+
+class Installer(object):
+    """Implements wheel installation logic
+
+    Supports Wheel version 1.0 (PEP 427).
+    """
+
+    def __init__(self, name, validators):
```
gaborbernat, line 29: We'll we need custom validators?

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+#
+# METADATA (and WHEEL) file parsing
+#
+def parse_metadata_file(contents):
+    # type: (Text) -> Message
+    feed_parser = FeedParser()
+    feed_parser.feed(contents)
+    return feed_parser.close()
```
gaborbernat, line 27: why return this?

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
...
+    "parse_metadata_file", "parse_record_file", "Hash", "Record", "InvalidRecord",
+]
+
+
+#
+# METADATA (and WHEEL) file parsing
+#
+def parse_metadata_file(contents):
+    # type: (Text) -> Message
+    feed_parser = FeedParser()
```
gaborbernat, line 25: shouldn't this be a context manager?

src/installer/parsers.py
pradyunsg: It returns the email.message.Message object that we want to be using as a return value of close.

src/installer/parsers.py
pradyunsg: No... This is a weird object, that's got kinda weird naming of it's API.

pradyunsg:
> I think we established that virtualenv needs finer grain control 👍 than the scope of this library, so it doesn't block me in any way.

Hmm... I'm not sure I follow. I thought that this would be able to cover the use case for virtualenv, since both source and destination are decoupled completely, and don't make assumptions about how the source is provided.

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+        # Ensure compatibility with this wheel version.
+        if metadata["Wheel-Version"] != "1.0":
+            raise InvalidWheelSource(
+                source, "Incompatible Wheel-Version: only support version 1.0",
+            )
+
+        self._pass_through_validators(source)
+
+        # Determine where archive root should go.
+        if metadata["Root-Is-Purelib"]:
```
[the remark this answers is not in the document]
pradyunsg: Could've. Doesn't hurt either way. 🤷🏻‍♂️

gaborbernat:
> Hmm... I'm not sure I follow. I thought that this would be able to cover the use case for virtualenv, since both source and destination are decoupled completely, and don't make assumptions about how the source is provided.

What do you define as a source? virtualenv does not install things straight from wheel... it rather generates pre-baked images on the disk and install is then essentially a copy/symlink. Furthermore, it still needs to trigger the console script generation in this case post copy/symlink. As I said this is highly custom and not sure we can/should force these restrictions on this library.

Additionally, virtualenv has probably the least aggressive python 2 deprecation policy, we'll probably support Python 2 until PyPy does it, and that's undefined for now.

FFY00:
> What do you define as a source? virtualenv does not install things straight from wheel... it rather generates pre-baked images on the disk and install is then essentially a copy/symlink.

You can subclass `Destination` and abstract all that, in fact, you are expected to. In my case, in python-install I unpack the wheel to the filesystem and work from there. Does this still not work in your use-case? If so, why?


> Furthermore, it still needs to trigger the console script generation in this case post copy/symlink. As I said this is highly custom and not sure we can/should force these restrictions on this library.

Script generation is optional and as such, should be provided independently from the install step. We might be able to come up with a model that works for you, but if we don't, you can still implement your own custom version.

pradyunsg:
> Then we can ask the user to define a get_full_path method, which could then be used by our rewrite_record. As an API user, I don't want to care about RECORD rewriting logic. This is also one of those things that people can easily get wrong, we want to prevent that.

Indeed!

I just spent a bit of time thinking, and I think it's possible to reduce Destination to a single method, by making write_file return the path that's supposed to go into the rewritten RECORD.

Since classes with a method are better represented as a function, I think we can reduce Destination down to a function.

FFY00:
> Since classes with a method are better represented as a function, I think we can reduce Destination down to a function.

I might want to save state, but in that case, I can implement a class with `__call__`.

pradyunsg:
> I just spent a bit of time thinking, and I think it's possible to reduce Destination to a single method, by making write_file return the path that's supposed to go into the rewritten RECORD.

It's not. :(

rewrite_record also serves as "the last operation". I do think that the earlier plan, to have a RecordKeeper class that Destination classes can use to keep track of how to write a RECORD is sufficient. Maybe, what we want to do is a mix of the two, with RecordKeeper being inside Installer, write_file returning the destination path, and being passed into rewrite_record?

pradyunsg:
> What do you define as a source?

For virtualenv, it'd be the prebaked image.

> install is then essentially a copy/symlink.

This can be what is done by the Destination object. Since the value from WheelSource.iter_files is passed straight into Destination.write_file, it's possible to use a object that holds the fioe descriptor, and performing the copy with it.

> Furthermore, it still needs to trigger the console script generation in this case post copy/symlink. 

This is completely controlled by the Destination class, which handles all the script generation.

> As I said this is highly custom and not sure we can/should force these restrictions on this library.

Thanks for elaborating on this. I think this API, as it stands, does accommodates for virtualenv's needs. I'd appreciate it if you could take a look at this, and see if/how that might not be the case.

gaborbernat:
> Thanks for elaborating on this. I think this API, as it stands, does accommodates for virtualenv's needs. I'd appreciate it if you could take a look at this, and see if/how that might not be the case.

I'd have to give a try to use it. I'll do it at some point but unlikely I'll have time this within the next week.

src/installer/_core.py
pradyunsg: We can have them. The RECORD matches the file hashes check is also a validator (i.e. opt in).

pradyunsg:
> RecordKeeper being inside Installer, write_file returning the destination path, and being passed into rewrite_record

I think I like this. It makes it straightforward enough for any of the consumers to depend on this library (return the paths from write_file, use the object passed into rewrite_record) while eliminating the need to determine file paths.

I do wonder how we'll map the hashes / sizes to the files, so that's certainly something for me to think about. I'll iterate on this with a piece of paper, later, to see what I can come up with.

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+        self.name = name
+        self._validators = validators
+
+    def install(self, source, destination, requested=False):
+        # type: (WheelSource, Destination, bool) -> None
+        """Install a wheel, as described by ``source``, to ``destination``.
+        """
+
+        # Process the WHEEL file
+        wheel_metadata_stream = source.open_dist_info(u"WHEEL")
```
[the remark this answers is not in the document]
pradyunsg: I felt that each step is really trivial and making them into private functions felt like an unnecessary strong tool.

The only thing I that thought could be useful is splitting the function into the "check WHEEL stuff" and "spread things out stuff". And, each of those is individually simple enough that it felt unnecessary to break them out into separate functions.

FWIW, I think the comments do a good enough job of describing what's happening, and the code flow is straightforward enough to follow.

We can definitely break this up if things get difficult to follow, but we're far from being in that position here. Maybe, adding RecordKeeper might be the reason to justify make the switch, since it'll add a bunch more complexity to the "spread stuff" part of this function.

(see https://github.com/pypa/pip/blob/master/src/pip/_internal/operations/prepare.py#L379 for what I think of as too-big-need-to-break).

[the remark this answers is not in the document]
gaborbernat: I always feel that when I need comments to tell what's happening and not why as a code smell and the architecture could be improved.

pradyunsg:
Cool, lemme know if we're not going to use this for virtualenv, since there might be simplifications possible if we're not going to be trying to accommodate for virtualenv's fairly different approach to dealing with things.

src/installer/_core.py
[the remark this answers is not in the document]
pradyunsg: I don't think I've ever thought of comments as being for describing "why" vs "how". Some food for thought.

FWIW, I'll be breaking this up when I add the record handling logic back into this class, so your point is well received.

agronholm:
> Maybe we should have some way to preserve stat(...).st_mtime from WheelSource to Destination?

If we don't, I see little point in recording the actual timestamps in the wheel file either.

FFY00:
> If we don't, I see little point in recording the actual timestamps in the wheel file either.

Reproducibility.

pradyunsg:
Fair enough. I'd imagine folks who care about build reproducibility definitely care.

100% open to suggestions on how to do this, since I'm actually not sure how to do this.

agronholm:
> Reproducibility.

How would it be less reproducible if we simply used a default timestamp, say `1970-01-01T00:00:00+00:00'?

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+    from installer._compat.typing import Text
+
+    Validator = Callable[[WheelSource], None]
+
+
+__all__ = ["Installer"]
+
+
+class Installer(object):
+    """Implements wheel installation logic
```
brettcannon, line 24:
```suggestion
    """Implements wheel installation logic.
```

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, name, validators):
+        # type: (Optional[Text], List[Validator]) -> None
+        super(Installer, self).__init__()
+        self.name = name
+        self._validators = validators
+
+    def install(self, source, destination, requested=False):
+        # type: (WheelSource, Destination, bool) -> None
+        """Install a wheel, as described by ``source``, to ``destination``.
+        """
```
brettcannon, line 38:
```suggestion
        """Install a wheel, as described by ``source``, to ``destination``."""
```

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Process the WHEEL file
```
brettcannon, line 40:
```suggestion
        # Process the WHEEL file.
```

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+    Supports Wheel version 1.0 (PEP 427).
+    """
+
+    def __init__(self, name, validators):
+        # type: (Optional[Text], List[Validator]) -> None
+        super(Installer, self).__init__()
+        self.name = name
+        self._validators = validators
+
+    def install(self, source, destination, requested=False):
```
brettcannon, line 35: What is "requested"? This could be interpreted as meaning, "someone explicitly requested the install". But this seems to be more about the `REQUESTED` metadata.

I would also say that should be a keyword-only parameter when Python 3-only support lands.

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Process the WHEEL file
+        wheel_metadata_stream = source.open_dist_info(u"WHEEL")
+        metadata = parse_metadata_file(wheel_metadata_stream)
+
+        # Ensure compatibility with this wheel version.
+        if metadata["Wheel-Version"] != "1.0":
+            raise InvalidWheelSource(
+                source, "Incompatible Wheel-Version: only support version 1.0",
+            )
```
brettcannon, line 48:
```suggestion
        wheel_version = metadata["Wheel-Version"]
        if  wheel_version != "1.0":
            raise InvalidWheelSource(
                source, "Incompatible Wheel-Version: version 1.0 supported, not {}".format(wheel_version),
            )
```
And this makes me miss the walrus operator. 😢

src/installer/_core.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self._pass_through_validators(source)
+
+        # Determine where archive root should go.
+        if metadata["Root-Is-Purelib"]:
+            root_scheme = Scheme.purelib
+        else:
+            root_scheme = Scheme.platlib
```
brettcannon, line 56:
```suggestion
        # Determine where archive root should go.
        root_scheme = Scheme.purelib if metadata["Root-Is-Purelib"] else Scheme.platlib
```

src/installer/_core.py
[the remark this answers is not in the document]
brettcannon: I actually find the one-liner a little easier to read as my brain distinguishes the 3 letter difference between `Scheme.purelib` and `Scheme.platlib` more in the ternary operator.

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
+"""Utilities for parsing and handling files related to a wheel (METADATA, WHEEL, RECORD)
+"""
```
brettcannon, line 2:
```suggestion
"""Utilities for parsing and handling files related to a wheel (METADATA, WHEEL, RECORD)."""
```

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
...
+    feed_parser.feed(contents)
+    return feed_parser.close()
+
+
+#
+# RECORD file parsing
+#
+class InvalidRecord(Exception):
+    """Raised when a Record is not valid, due to improper element values or count.
+    """
```
brettcannon, line 35:
```suggestion
class InvalidRecord(Exception):

    """Raised when a Record is not valid, due to improper element values or count."""
```

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
...
+        self.value = value
+
+    def __repr__(self):
+        return "Hash(name={!r}, value={!r})".format(self.name, self.value)
+
+    @classmethod
+    def parse(cls, h):
+        # type: (str) -> Hash
+        name, value = h.split("=", 1)
+        return Hash(name, value)
```
brettcannon, line 61:
```suggestion
    def parse(cls, specified_hash):
        # type: (str) -> Hash
        name, value = specified_hash.split("=", 1)
        return Hash(name, value)
```

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
...
+            except ValueError:
+                issues.append("`hash` does not follow the required format")
+        else:
+            hash_value = None
+
+        if size:
+            try:
+                size_value = int(size)  # type: Optional[int]
+            except ValueError:
+                issues.append("`size` cannot be non-integer")
```
brettcannon, line 114:
```suggestion
                issues.append("`size` cannot be a non-integer")
```

src/installer/parsers.py
```diff
@@ -0,0 +1,137 @@
...
+        if issues:
+            raise InvalidRecord(elements=(path, hash_, size), issues=issues)
+
+        return cls(path=path, hash_=hash_value, size=size_value,)
+
+
+def parse_record_file(rows):
+    # type: (Iterator[str]) -> Iterator[Record]
+    """Parse a RECORD file, provided as an iterator of record lines.
+    """
```
brettcannon, line 127:
```suggestion
    """Parse a RECORD file, provided as an iterator of record lines."""
```

src/installer/protocols.py
```diff
@@ -0,0 +1,120 @@
...
+    from installer._compat.typing import Text, FSPath
+
+
+Scheme = Enum("Scheme", ["purelib", "platlib", "headers", "scripts", "data"])
+
+
+@with_metaclass(abc.ABCMeta)
+class WheelSource(object):
+    """Represents an installable wheel.
+    """
```
brettcannon, line 20:
```suggestion
    
    """Represents an installable wheel."""
```

src/installer/protocols.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, distribution, version):
+        # type: (Text, Text) -> None
+        super(WheelSource, self).__init__()
+        self.distribution = distribution
+        self.version = version
+
+    @property
+    def dist_info_dir(self):
+        """Name of the dist-info directory.
+        """
```
brettcannon, line 31:
```suggestion
        """Name of the dist-info directory."""
```

src/installer/protocols.py
```diff
@@ -0,0 +1,120 @@
...
+        return u"{}-{}.dist-info".format(self.distribution, self.version)
+
+    @property
+    def data_dir(self):
+        """Name of the data directory.
+        """
```
brettcannon, line 37:
```suggestion
        """Name of the data directory."""
```

src/installer/protocols.py
```diff
@@ -0,0 +1,120 @@
...
+
+            >>> f = wheel_source.open_dist_info("INSTALLER")
+            >>> f.read()
+            'pip'
+        """
+        raise NotImplementedError()
+
+    # All files in the wheel
+    @abc.abstractmethod
+    def iter_files(self):
```
brettcannon, line 67: I will say I dislike `iter_*()` API names as it doesn't communicate what the method actually does, just what you are expected to do with it or what the turn type is depending if you read that as shorthand as `iterate` or `iterator`, respectively. I personally would be fine with `files()` and simply document it returns an iterator.

src/installer/protocols.py
```diff
@@ -0,0 +1,120 @@
...
+        # type: () -> Iterator[Tuple[FSPath, BufferedReader]]
+        """Sequential access to all contents of the wheel (including dist-info files).
+
+        All paths must be posix-style paths (i.e. can be used with ``posixpath`` module)
+        relative to the root of the wheel.
+
+        Sample usage/behavior::
```
brettcannon, line 74:
```suggestion
        Sample usage/behaviour::
```
😁

src/installer/protocols.py
```diff
@@ -0,0 +1,120 @@
...
+
+            >>> iterable = wheel_source.iter_files()
+            >>> next(iterable)
+            ('pkg-1.0.0.dist-info/METADATA', <...>)
+
+        This method may be called multiple times, and each iterator returned must
+        provide the same content upon reading from the streams.
+        """
+        raise NotImplementedError()
```
brettcannon, line 83:
```suggestion
        raise NotImplementedError
```

src/installer/_core.py
pradyunsg: That interpretation is correct, since that's what the REQUESTED file represents.

I haven't added any details to this docstring yet, so I'll document the parameter when I get to this.

OK to make it a keyword-only parameter (will document so in the initial version as well).

src/installer/protocols.py
pradyunsg: I thought this was necessary for Python 2?

src/installer/_core.py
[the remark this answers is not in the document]
pradyunsg: I don't really mind either form, and if y'all reckon it's better, I'm happy to change over. :)

src/installer/protocols.py
uranusjr: It’s not, Python 2.7 can raise from exception class. (Although fwiw I personally like to instantiate the exception anyway.)

src/installer/parsers.py
pradyunsg: That'll make the line too long. :)

src/installer/protocols.py
```diff
@@ -0,0 +1,118 @@
...
+from installer._compat.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from io import BufferedReader
+    from typing import List, Sequence, Tuple
+
+    from installer._compat.typing import Text, FSPath
+
+
+Scheme = Enum("Scheme", ["purelib", "platlib", "headers", "scripts", "data"])
```
uranusjr, line 14: Do we need this to be an `Enum`? The available scheme keys are [well-defined in `sysconfig`](https://docs.python.org/3/library/sysconfig.html#installation-paths), and I imagine users are going to use `sysconfig.get_paths()` in the vast majorities of cases anyway. I feel this is not enough reason to pull in an extra dependency for Python 2. If type safty is the concern, `typing.Literal` would work well enough IMO.

src/installer/protocols.py
uranusjr: Personally I like having `iter_` since the fact that an iterator is returned is the second most important part of what the function does (the first is it iterators through files). Python does not offer any way to guard against iterator reuse, and the only thing to prevent accidental usage is to stick something into the name. Type-checking does not help here because it would happily accept usages like

```python
def install_files(files: Iterable[File]):
   ...

files = source.files()
logger.debug('Processing %s', [path for path, _ in files])
install_files(files)
```

I won’t speak for others, but I can’t tell whether this code is correct or not in a quick code review. This works if `files()` returns a list, but not if it returns an iterator. Maybe I should call `files()` twice? But that wouldn make things worse if the function implies significant performance impact. I can go through this thought process every time I see this, and potentially dig into the documentation or source code. Or I can easily avoid all the mental work by simply adding `iter_` to the function name, with a slight aesthetics cost. The trade-off is worth it IMO.

src/installer/protocols.py
```diff
@@ -0,0 +1,118 @@
...
+
+        Example usage/behavior::
+
+            >>> dest.write_file(Scheme.purelib, "__init__.py", stream1)
+            >>> dest.write_file(Scheme.purelib, "awesome_wrapper.py", stream2)
+            >>> dest.write_file(Scheme.purelib, "_awesome.pyd", stream3)
+            >>> dest.rewrite_record(Scheme.purelib)
+
+        """
+        raise NotImplementedError()
```
uranusjr, line 118: It’s probably worthwhile to also implement the most common scenario as a concret class:

```python
class FilesystemDestination(Destination):
    def __init__(self, scheme_paths: Dict[Scheme, FSPath]):
        self._scheme_paths = scheme_paths

    def write_file(self, scheme, path, stream):
        with io.open(os.path.join(self._scheme_paths[scheme], path), 'wb') as f:
            f.write(stream.read())

    def rewrite_record(self, scheme):
        ...
```

but this can be implemented as a follow-up PR.

pradyunsg: Yea, that's the plan, to do so in a follow up PR.

src/installer/protocols.py
pradyunsg: We don't. I'm 100% OK to change this. :)

src/installer/protocols.py
pradyunsg: Alrighty! Will remove! ^>^

src/installer/protocols.py
pradyunsg: I'm gonna have to change this for timestamp + record-keeping reasons.

For now, I'm think gonna call this `get_contents`, returning a `Tuple[Record, BufferedReader, Optional[stat_result]]`. I'm a bit concerned with using `Record` here directly, since @agronholm has noted that he wants to be able to make `wheel`'s `WheelFile` capable of being a `WheelSource`.

agronholm:
@pradyunsg So, how would you prefer us to align the APIs? The new WheelFile (and related) APIs are going to be Python 3 only, while this project seems to be design with Python 2 compatibility for reasons I don't really understand. This makes it impossible for you to import my code directly.

src/installer/protocols.py
agronholm: So are we going to restore the modification time and file attributes present in the archive when the wheel is being installed? Sadly the wheel PEP does not say anything about this which I feel is a significant problem.

pradyunsg:
FWIW, I think `wheel.WheelFile` can just implement the `WheelSource` protocol and be usable with this library. The `WheelSource` API (by design) permits Python 3 users to use `pathlib.Path` if they want to, and `wheel.WheelFile` would by a Py3 only `WheelSource`.

The main blocker is that I need to still add 2 things here:

- "`RECORD` management" in `Installer`, so that `rewrite_record` gets a `RecordKeeper` object in addition to the scheme.
- Add RECORD-entry + optional timestamp information in `iter_files` (which should probably get renamed).

I don't want us to end up introducing a wheel -> installer dependency, so the `WheelSource` protocol shouldn't depend on anything else from `installer`. Gimme a day or so to iterate here, and figure out what's the best way to add these things, without making it difficult to implement the `WheelSource` protocol in `wheel`.

src/installer/protocols.py
pradyunsg: Looking at [pip's logic](https://github.com/pypa/pip/blob/acab2ee54ed25edd44edbdf793a4b1d173bfbcb1/src/pip/_internal/operations/install/wheel.py#L665), we unpack the zip with the following guarentees:

> All files are written based on system defaults and umask (i.e. permissions are not preserved), except that regular file members with any execute permissions (user, group, or world) have "chmod +x" applied after being written. Note that for windows, any execute changes using os.chmod are no-ops per the python docs.

And then install from the unpacked wheel with, among other things:

```py
                # Copy over the metadata for the file, currently this only
                # includes the atime and mtime.
                st = os.stat(srcfile)
                if hasattr(os, "utime"):
                    os.utime(destfile, (st.st_atime, st.st_mtime))
```

I'm very confused right now about what the potential details we'd need, since IIUC, pip's not preserving the details during unpacking, but preserving them when moving the files (!).

pradyunsg: I have so many questions:

- Can we have empty folders in a wheel? If yes, how should they get installed?
- What metadata about a file in a wheel, do we actually care about?
	- Executable bits?
	- What properties from `os.stat` / `ZipInfo`?
- Would people be OK with files being written with timestamp=0? (this would be a concrete Destination implementation thing, so I'm not too worried)

src/installer/records.py
```diff
@@ -130,8 +145,7 @@ def from_elements(cls, path, hash_, size):
 
 def parse_record_file(rows):
     # type: (Iterator[str]) -> Iterator[Record]
-    """Parse a RECORD file, provided as an iterator of record lines.
-    """
+    """Parse a RECORD file, provided as an iterator of record lines."""
     reader = csv.reader(rows, delimiter=",", quotechar='"', lineterminator=os.linesep)
```
agronholm, line 149: `lineterminator` should not depend on the OS on which the library is used! The wheel library always uses `\n` as the line terminator, regardless of the operating system.

src/installer/protocols.py
pfmoore: The answers to these questions should be added to the wheel spec, by the way 🙂

src/installer/records.py
pfmoore: This is also a detail that should be added to the wheel spec.

Just to be clear, all I'm saying is that these questions come up whenever anyone works on a wheel implementation, and we should capture them and make sure there's a clear answer recorded somewhere - and the spec is probably the best place.

pradyunsg: https://www.python.org/dev/peps/pep-0376/#record

PEP 376 says use os.linesep.

pradyunsg: @pfmoore can/should I move things over to packaging.python.org in the process?

uranusjr: PEP 427 (the wheel spec) simply defers to PEP 376 (which defines `.dist-info` directories) on `RECORD`, but that spec can afford to be platform-dependant (IMO it does not need to, but that’s another issue), while the wheel format shouldn’t be. From what I can read from the wheel spec, there are two things in `RECORD` that may need clarification.

The first is `lineterminator`:

> The `RECORD` file is a CSV file, composed of records, one line per installed file. The `csv` module is used to read the file, with these options:
>
> * field delimiter : `,`
> * quoting char : `"`
> * line terminator : `os.linesep` (so `\r\n` or `\n`)

and the other is the third path element variant:

> the file's **path**
> * a `'/'`-separated path, relative to the base location, if the file is under the base location.
> * a `'/'`-separated path, relative to the base location, if the file is under the installation prefix AND if the base location is a subpath of the installation prefix.
> * an absolute path, using the local platform separator.

Both values should be interpreted in a platform-dependent way, which does not work well for wheels.

My proposal is to adding the followings to PEP 427 (item 4 in [this section](https://www.python.org/dev/peps/pep-0427/#the-dist-info-directory)):

> The `RECORD` file […]. Unlike PEP376 […]. The file should always be read using filed delimiter `,`, quoting char `"`, and line terminator `\n`. The **path** component should always use `'/'` as the separator, instead of the local platform path separator.

I can draft a pull request to the PEP repository if this makes sense.

pradyunsg: I'll change lineterminator to "\n". :)

src/installer/protocols.py
uranusjr: > Can we have empty folders in a wheel? If yes, how should they get installed?

No. [There is a Dicourse thread proposing to amend this.](https://discuss.python.org/t/refine-wheel-spec-to-designate-directory-entries/1667)

> Would people be OK with files being written with `timestamp=0`? 

This would cause pain when integrating wheels into timestamp-based incremental build systems (e.g. GNU Make). I think setting timestamp to 0 would be worse than what pip currently does (which sets the timestamp to whenever the wheel is installed, I believe?)

src/installer/records.py
pradyunsg: @pfmoore looking at https://discuss.python.org/t/refine-wheel-spec-to-designate-directory-entries/1667 and https://github.com/python/peps/pull/1045, I wonder if we should make these changes in-place to wheel 1.0, or if they should be in a wheel 1.1.

I'm 100% OK either way, with a slight preference toward the former since we're amending a standard based on current "relevant" practice in all the tools we know of.

src/installer/protocols.py
pradyunsg: > > Can we have empty folders in a wheel? If yes, how should they get installed?
>
> No

Well, I'm gonna implement status quo, so not implementing any support for this sounds good. :)

> timestamp-based incremental build systems

To be clear, the situation would be that WheelSource objects wouldn't provide timestamp information, so Destination objects would set the timestamp based on whatever logic they want - I imagine pip would preserve its current approach and keep the time dependent on unpack time.

The question really should've been, do we care about preserving timestamps from a WheelSource when installing from it? I think the answer is no, so I'm going to move forward with that assumption.

src/installer/records.py
pfmoore: > PEP 427 (the wheel spec) simply defers to PEP 376 (which defines .dist-info directories) on RECORD

Actually, it doesn't. At no point that I could find does it say that the RECORD file in the wheel follows the format of the PEP 376 RECORD file. (It *nearly* says it in a few places, but never actually comes out and does so). And PEP 376 is problematic in other ways - for example, it allows absolute filenames, which make no sense in a wheel. Longer term, I think PEP 427 is badly under-specified and needs a revision to fix the various "open to interpretation" problems it has. That rewrite would need to be a "Wheel 1.1" spec.

However, in the interest of not derailing momentum on *this* project, I think we can take the following view:

* `RECORD` should use `,` as delimiter, `"` as quote, and `\n` as line separator.
* `RECORD` should use `/` for paths, which must all be relative to the wheel root.

I'd be OK with these two changes being published as an update to the Wheel 1.0 spec. But they would need a note covering compatibility. Something like "the spec was never intended to allow platform-dependent line and path separators in the `RECORD` file, but the original wording did not make that clear. As no known tools ever took advantage of the ability to do this, the spec has now been changed to explicitly disallow it".

The change should be posted to the Packaging Discourse before being submitted, just to make sure the packaging community has a chance to comment. I don't expect it to be controversial, but hey, this is packaging 🙄 

I'm trying really hard to avoid needing a Wheel 1.1 here, because if we decide we need Wheel 1.1 to clear up the vagueness in 1.0, we have a hard decision to make about how this project will handle files that declare they follow Wheel 1.0 (i.e., every existing wheel...)

By the way, while we're discussing `RECORD` files, in theory when this project installs a wheel, it should write a `RECORD` file in PEP 376 format, *not* in PEP 427 format. In other words, os.linesep and the weird `/`-if-relative-else-platform-native filenames. I'm betting we won't do that. But if not, we should probably tidy up PEP 376 to match what tools really do, as well.

(Sigh. Remember the good old days when PEPs could get away with being hand-wavey and vague? I think I need to go and give past me a really good talking to...)

pfmoore: > @pfmoore can/should I move things over to packaging.python.org in the process?

I think that making packaging.python.org the canonical spec itself needs a PEP - see how the core metadata specs did it.

src/installer/protocols.py
agronholm: To me it seems correct to leave the timestamp in newly created files alone. If pip indeed does this already, then that opens the door for wheel to use the zip default timestamp for all files (instead of copying the timestamp from the file system). This too should be clarified in the new PEP which @uranusjr is apparently writing now.

src/installer/records.py
pradyunsg: > I'm trying really hard to avoid needing a Wheel 1.1 here,

I agree -- let's just amend wheel 1.0 based on "what actually happens" with the suggested wording for the note.

> I think that making packaging.python.org the canonical spec itself needs a PEP - see how the core metadata specs did it.

I don't think we don't *need* a PEP -- it's just been a convenient occasion to make the switch to packaging.python.org along with a PEP for updates (I remember @ncoghlan has stated so in a bunch of places, but I'll need to go digging to find one of these statements).

FWIW, we're making progress on this project + wheel's API toward better quality implementation / tooling for handling + installing wheels; as well as amending the wheel 1.0 PEP. I think that's as an time as any, to make the switch. :)

pfmoore: If we move the spec to packaging.python.org (I'd be interested if you could find what @ncoghlan said - he was on the SC so has a better idea of how far we can mangle the standard PEP processes than I do 😉) then I'd prefer if we tidied it up as part of that exercise, and didn't just copy it over as is. There's a chunk of tidying up needed¹ and the move would be the best time to do that.

If you're thinking of just moving the text across unchanged (or with the sort of minor tweaks we've discussed here) I don't really see the point, TBH. It's not like a PR against the PEP is any harder than a PR against the packaging docs... (most of this discussion is about moving the docs, not about the changes which I already said were OK as a PEP modification)

¹ For example, I just spotted that the only discussion of hashes in `RECORD` is in the "Signed wheels" section, even though signed wheels are hardly ever used!

agronholm: > For example, I just spotted that the only discussion of hashes in RECORD is in the "Signed wheels" section, even though signed wheels are hardly ever used!

Maybe that's because without signing, `RECORD` does not protect against tampering, as anybody who wanted to modify the wheel could simply edit `RECORD` to match.

pradyunsg: https://github.com/pradyunsg/installer/pull/20 covers this. All the other discussion is preserved, but as @pfmoore noted, isn't 100% on-topic for this project/PR. :)

pradyunsg:
Note to self: need to allow for some mechanism to add "install-time metadata" like INSTALLER, REQUESTED, direct_url.json (PEP 610) and more in the future (like proposed [HASH](https://discuss.python.org/t/draft-pep-recording-the-source-hash-of-installed-distribution/4660) file) without hard-coding support for the specific files in the code. This would allow for addition of these files to be done transparently as the .dist-info directory's contents evolve.

brettcannon:
What's next on this PR (and project)? More time? 😉

pradyunsg:
> More time?

Yea, pretty much.

ofek:
Any update?

pradyunsg:
I've blocked out a few hours in the coming week to pull this forward, including a small chunk of co-working time with @FFY00! :)

pradyunsg:
A few notes:

- I can't find the notes from my and @FFY00's conversation. I'd like to blame my broken MacBook. :)
- there's a few tweaks to be made to the API at play here, and there's a need for some wrangling w.r.t. handling RECORD entries properly in this.
- I have an open offer for folks to co-work with me on this (https://discuss.python.org/t/pep-517-workflow-for-distributions/5959/5?u=pradyunsg) which would help me concentrate on getting this over-the-line.
- There are abstract interfaces that the "core" `Installer` class would see, and those would need concrete implementations. I expect the abstract interfaces to be subject to change until we hit "stable" on this project.
- I would still like downstream folks to try to integrate with this -- I'll communicate more about this on #1, once this is ready-enough to actually try using. :)

pradyunsg:
Update: @lkisskol and I are working on getting the ball rolling again this week.

pradyunsg:
Closing since this isn't going to be the primary PR where work is done. Please follow #1 instead. :)

PR #22, opened 2020-12-15
Provide new implementation for copying files

lkollar:
This provides the IOHandler class (name not final) to perform I/O operations for installing wheels.

### DIFF
--- a/src/installer/iohandler.py
+++ b/src/installer/iohandler.py
@@ -1,0 +1,121 @@
+"""Handling wheel installation I/O-related functionality."""
+import contextlib
+import hashlib
+import os
+
+from installer._compat.typing import TYPE_CHECKING
+
+from .records import Hash, Record
+
+if TYPE_CHECKING:
+    from typing import BinaryIO, Callable, Dict, Iterator, Tuple
+
+    from ._compat.typing import FSPath
+
+    CopyHandler = Callable[[BinaryIO, BinaryIO, str], Tuple[int, Hash]]
+
+_BLOCK_SIZE = 4096
+
+
+def basic_copy_handler(
+    source,  # type: BinaryIO
+    dest,  # type: BinaryIO
+    hash_algorithm="sha256",  # type: str
+):
+    # type: (...) -> Tuple[int, Hash]
+    """Copy a buffer and compute its hash and size.
+
+    Copies the source buffer into the specified destination buffer. It also computes
+    the hash of the file to avoid traversing the source twice.
+
+    :param source: buffer holding the source data
+    :param dest: destination buffer
+    :param hash_algorithm: hashing algorithm
+
+    :return: size and ``Hash`` tuple of the copied buffer
+    """
+    hasher = hashlib.new(hash_algorithm)
+    size = 0
+    while True:
+        buf = source.read(_BLOCK_SIZE)
+        if not buf:
+            break
+        hasher.update(buf)
+        dest.write(buf)
+        size += len(buf)
+
+    hash_ = Hash.parse("sha256={}".format(hasher.hexdigest()))
+    return size, hash_
+
+
+class IOHandler(object):
+    """Implement I/O operations for installers."""
+
+    def __init__(
+        self,
+        scheme_path_map,  # type: Dict[str, FSPath]
+        hash_algorithm="sha256",  # type: str
+        copy_handler=basic_copy_handler,  # type: CopyHandler
+    ):
+        # type: (...) -> None
+        """Construct an ``IOHandler`` instance.
+
+        :param scheme_path_map: mapping of scheme names to paths
+        :param hash_algorithm: hashing algorithm passed to the ``copy_handler``
+        :param copy_handler: callable implementing the copy and hashing operation
+        """
+        self._scheme_path_map = scheme_path_map
+        self._hash_algorithm = hash_algorithm
+        self._copy_handler = copy_handler
+
+    @contextlib.contextmanager
+    def open_source(
+        self,
+        scheme,  # type: str
+        source_path,  # type: FSPath
+    ):
+        # type: (...) -> Iterator[BinaryIO]
+        """Open the specified source and provide a destination buffer.
+
+        This function will determine the correct destination based on the specified
+        scheme and provide a buffer where the contents can be written.
+
+        :param scheme: scheme of the source
+        :param source_path: relative path to the file in the wheel
+
+        :return a buffer for the destination
+        """
+        try:
+            dest_dir = self._scheme_path_map[scheme]
+        except KeyError:
+            raise ValueError("No such scheme: {}".format(scheme))  # FIXME
+
+        dest_path = os.path.join(dest_dir, source_path)
+        parent_path = os.path.dirname(dest_path)
+
+        if not os.path.exists(parent_path):
+            # FIXME raise custom exception
+            os.makedirs(parent_path)
+
+        with open(dest_path, "wb") as dest_buf:
+            yield dest_buf
+
+    def copy_file(
+        self,
+        scheme,  # type: str
+        path,  # type: FSPath
+        stream,  # type: BinaryIO
+    ):
+        # type: (...) -> Record
+        """Copy the provided path into the appropriate scheme location.
+
+        :param scheme: destination scheme
+        :param path: path inside the Python package
+        :param stream: source file contents
+
+        :return record of the installed file
+        """
+        with self.open_source(scheme, path) as dest_buf:
+            size, hash_ = self._copy_handler(stream, dest_buf, self._hash_algorithm)
+
+        return Record(path, hash_=hash_, size=size)
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -58,6 +58,18 @@ class Hash(object):
         # type: () -> str
         return "Hash(name={!r}, value={!r})".format(self.name, self.value)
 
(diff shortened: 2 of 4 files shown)


src/installer/records.py
```diff
@@ -111,6 +123,22 @@ def __repr__(self):
             self.path, self.hash_, self.size,
         )
 
+    def __eq__(self, other):
```
pradyunsg, line 126: We'd want to add testa for these as well.

lkollar: Good point, will do.

src/installer/records.py
```diff
@@ -111,6 +123,22 @@ def __repr__(self):
...
+        # type: (object) -> bool
+        if not isinstance(other, Record):
+            return NotImplemented
+        return (
+            self.path == other.path
```
FFY00, line 131: Isn't the relevant part here the path? I feel like if we have multiple records with the same path but different hashes/size there is probably something wrong, I would raise and exception.

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+from installer._compat.typing import TYPE_CHECKING
+
+from .records import Hash, Record
+
+if TYPE_CHECKING:
+    from ._compat.typing import FSPath
+    from typing import Dict, Tuple, Optional, Callable
+
+_BLOCK_SIZE = 4096
+_SCHEME_KEYS = ["platlib", "purelib"]
```
FFY00, line 16: This is not being used anywhere. Leftover?

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+    Copies the source buffer to the specified destination. It also computes the hash
+    of the file to avoid traversing the source twice.
+
+    :param source: buffer holding the source data
+    :param dest: destination path
+    :param hash_algorithm: hashing algorithm
+
+    :return: size and ``Hash`` tuple of the copied buffer
+    """
+    assert hash_algorithm is not None  # to please mypy
```
FFY00, line 36: Why are we using an assert? mypy is right in calling it out, we shouldn't use an assert (which can be disabled with `-OO` :stuck_out_tongue:) just to please mypy. I think we should raise an exception.

FFY00 (changes requested):
Just some (mostly) minor comments.

I assume you and @pradyunsg discussed the API interface, so I am not gonna comment on that.

Thanks :blush:

src/installer/iohandler.py
uranusjr: Why is `hash_algorithm` allowed to be None in the first place? What does it mean?

FFY00: That's API design I did not question it as I said the in general review commentary. I assumed there was a reason this could be `None` and given the assert, I suggested bumping it up to a proper exception.

But looking into it, `RECORD` does not have a hash algorithm and still needs to be copied. We should just skip the hash checking if `hash_algorithm` is `None` instead of exploding.
Breaking PEP 427, but happening in the wild, some backends are not generating hashes for `.pyc` files.
This seems to include setuptools, which is problematic. It is a bit late today, I will see if tomorrow I can look into it.

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+    from typing import Dict, Tuple, Optional, Callable
+
+_BLOCK_SIZE = 4096
+_SCHEME_KEYS = ["platlib", "purelib"]
+
+
+def basic_copy_handler(
+    source,  # type: RawIOBase
+    dest,  # type: FSPath
+    hash_algorithm="sha256",  # type: Optional[str]
```
pradyunsg, line 22: This shouldn't be `Optional`.

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+
+if TYPE_CHECKING:
+    from ._compat.typing import FSPath
+    from typing import Dict, Tuple, Optional, Callable
+
+_BLOCK_SIZE = 4096
+_SCHEME_KEYS = ["platlib", "purelib"]
+
+
+def basic_copy_handler(
```
pradyunsg, line 19: Let's chat about the naming in our call today. :)

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class IOHandler(object):
+    """Implement I/O operations for installers."""
+
+    def __init__(
+        self,
+        scheme_path_map,  # type: Dict[str, FSPath]
+        hash_algorithm="sha256",  # type: Optional[str]
+        copy_handler=basic_copy_handler,  # type: Callable[[RawIOBase, FSPath, Optional[str]], Tuple[int, Hash]]
```
pradyunsg, line 59: I'd make this type aliased in the `if TYPE_CHECKING`, via `CopyHandler = ...` assignment there.

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+    :param source: buffer holding the source data
+    :param dest: destination path
+    :param hash_algorithm: hashing algorithm
+
+    :return: size and ``Hash`` tuple of the copied buffer
+    """
+    assert hash_algorithm is not None  # to please mypy
+    hasher = hashlib.new(hash_algorithm)
+    size = 0
+    with open(dest, "wb") as dest_fd:
```
pradyunsg, line 39: I'd prefer we open the file in the `IOHandler`, so that this function isn't tied to the filesystem.

src/installer/iohandler.py
```diff
@@ -0,0 +1,100 @@
...
+            dest_dir = self._scheme_path_map[scheme]
+        except KeyError:
+            raise ValueError("No such scheme: {}".format(scheme))  # FIXME
+
+        dest_path = os.path.join(dest_dir, path)
+        parent_path = os.path.dirname(dest_path)
+
+        if not os.path.exists(parent_path):
+            # FIXME raise custom exception
+            os.makedirs(parent_path)
```
pradyunsg, line 97: I'd factor this out into a method, so that downstream could override it, and change behaviors?

src/installer/iohandler.py
pradyunsg: Yea, looks like. We'd probably want to use `SCHEME_KEYS` from `installer.utils`.

src/installer/records.py
lkollar: I don't think it's a good idea to raise an exception while checking for equality. I would find it surprising as a user. As an equality check I don't think we should make assumptions about how it will be used. Consider a case where we rewrite the shebang in a script: if one builds a record of the original file and one for the installed file, the records will have different hashes and likely sizes as well. These should just compare as not equal in my opinion.

src/installer/iohandler.py
lkollar: Yup, good catch!

pradyunsg (changes requested):

src/installer/iohandler.py
lkollar: I wonder if we should just defer this to the copy handler. That way that implementation can decide how they want to deal with paths. For example, an (admittedly weird) implementation which wants to install wheels into a database could leave everything intact in `IOHandler` and only supply a different copy handler to achieve this.

src/installer/iohandler.py
lkollar: But then IOHandler will be tied to the FS. Although `os.makedirs` ties it to it too, but I want to move that into the copy handler as well.

pradyunsg: I'd rather have that be tied to the file system in a specific method within it, rather than this function by tied to the file system -- this can directly consume streams, which is useful when you're transforming the contents of the wheel (eg: shebang rewriting).

pradyunsg:
> I assume you and @pradyunsg discussed the API interface, so I am not gonna comment on that.

FYI: this is still under active discussion, so thoughts are very welcome on this.

src/installer/iohandler.py
FFY00: What if the files is not supposed to have an hash? Which is the case of `RECORD`.

src/installer/records.py
```diff
@@ -111,6 +123,22 @@ def __repr__(self):
             self.path, self.hash_, self.size,
         )
 
+    def __eq__(self, other):
+        # type: (object) -> bool
+        if not isinstance(other, Record):
+            return NotImplemented
```
FFY00, line 129: Did you mean `raise` here?

FFY00 (changes requested):

src/installer/records.py
FFY00: I appreciate the semantics we get from just having this return `False`. As you explain it does make sense from a logical/structural POV. But I also think the benefits of having this explode on unwanted cases heavily outweight that.

I feel like the current API induces users to wrongly use it. But I guess it comes down to how you promote it being used.
In which use cases would you use this? Or rather, which problem does this solve?
(I wait for your response to proceed the discussion in order to avoid bikesheding :sweat_smile:)

src/installer/records.py
lkollar: No, this [should be returned](https://docs.python.org/3.9/library/constants.html#NotImplemented) AFAIK.

pradyunsg: No -- returning here is fine: https://docs.python.org/3/library/constants.html#NotImplemented

FFY00: Ah, sorry.

PR #23, opened 2020-12-16
Add a string representation to `Record`

pradyunsg:
* Add `__str__` to `installer.records.Record` and `installer.records.Hash`.
* Add tests for the string representation


pradyunsg:
Merging despite failing CI because this + #24 fixes the linter failure.

Tested locally that this passes tests on MacOS. If it breaks on Windows/Ubuntu, we can iterate on this. Comments welcome, as always. :)

PR #25, opened 2020-12-16
Rework CI to better utilise available workers

pradyunsg:


pradyunsg:
4m 17s to 1m 44s, despite MacOS having longer queue times than usual.

PR #26, opened 2020-12-17
Implement the parsing utilities for end users

pradyunsg:
TODO:

- [x] tests
- [x] code
- [x] docs

### DIFF
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -1,0 +1,70 @@
+"""Utilities related to handling / interacting with wheel files."""
+
+import os
+import re
+from collections import namedtuple
+from email.parser import FeedParser
+
+from installer._compat.typing import TYPE_CHECKING, cast
+
+if TYPE_CHECKING:
+    from email.message import Message
+    from io import BufferedReader
+    from typing import Any, Callable, Dict, Iterable, NewType, Tuple
+
+    from installer._compat.typing import Binary, Text
+
+    Scheme = NewType("Scheme", str)
+    AllSchemes = Tuple[Scheme, ...]
+
+__all__ = [
+    "parse_metadata_file",
+    "parse_wheel_filename",
+    "WheelFilename",
+    "SCHEME_NAMES",
+]
+
+_WHEEL_FILENAME_REGEX = re.compile(
+    r"""
+    ^
+    (?P<distribution>.+?)
+    -(?P<version>.*?)
+    (?:-(?P<build_tag>\d[^-]*?))?
+    -(?P<tag>.+?-.+?-.+?)
+    \.whl
+    $
+    """,
+    re.VERBOSE | re.UNICODE,
+)
+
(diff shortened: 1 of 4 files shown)


src/installer/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    AllSchemes = Tuple[Scheme, ...]
+
+__all__ = [
+    "parse_metadata_file",
+    "parse_wheel_filename",
+    "WheelFilename",
+    "SCHEME_NAMES",
+]
+
+_WHEEL_FILENAME_REGEX = re.compile(
```
brettcannon, line 25: Does this mean you are not taking a dependency on 'packaging'? If not then can we take a dependency on 'installer' in 'packaging'? That parsing code will very likely be needed in 'packaging' when we add metadata support there.

brettcannon: I.e. 'packaging' has https://packaging.pypa.io/en/latest/tags.html#packaging.tags.parse_tag to parse while tags. If you want a literal `parse_wheel_filename()` or `parse_wheel_path()` class method we could add that to `packaging.tags`.

FFY00: I would heavily prefer to not have a dependency on pypa/packaging here as it would become a bootstrapping dependency.

pypa/packaging would be required to check for if the wheel dependencies are available, but I think we should have a `--skip-dependencies` flag like pypa/build to simply install the wheel without caring. In that case we would still probably want to parse the wheel name and depening on pypa/packaging for that function would make it a bootstrapping dependency.

pradyunsg: I think in the _very_ short term, this is better -- I don't think these exist anywhere else in a reusable form, and I'd prefer to use them internally anyway. :)

Once packaging gets the metadata utilities, I think I'll be happy to add a dependency on `installer` -> `packaging` and move some things there; assuming we're also able to drop pyparsing, due to the import time cost of using it.

pradyunsg:
(merging, not because I think there's no more discussion to have here, but because I'd like to make other changes that depend on this -- PRs are cheap)

src/installer/utils.py
brettcannon: So then are we expecting 'packaging' to take a dependency on 'installer'? Either way, writing the same code twice is going to lead to potential semantic skewing that I want to avoid, so someone is going on somebody else somehow.

But do note there are already dependencies here  on importlib-resources so 'installer' is not dependency-free ATM.

pradyunsg: > But do note there are already dependencies here on importlib-resources so 'installer' is not dependency-free ATM.

Oh yea, but that isn't an intentional goal for me. It's just that it's easy enough to implement these utilities without dependencies. :)

At this point, I'm only concerned about making things work properly -- I'd like to have the conversation on what belongs where and all that once the initial round of implementation is done.

PR #28, opened 2020-12-18
Add initial "source" and "destination" abstractions

pradyunsg:
A couple of rough design notes:

- The `WheelSource` is intentionally not tied to any properties within this package -- this allows other projects (such as `wheel`) to implement this interface without needing to add a dependency.
- The `WheelDestination` is where nearly all of the logical "nuance" of script generation and RECORD writing lives. This is an intentional choice. This allows for the "core" to only need to know how the wheel file is structured, and correctly invoke this class.

This is NOT "the final API" -- the plan/expectation is that there would likely be more iterations on this as we get feedback on how this works when trying implementing this interface; as the "core" is implemented and as the concrete implementations for these are written. And, yes, there's also gonna be actual useful concrete implementations of these abstractions, that'll live right next to them. Those are what I'd expect normal humans to use.

But one thing at a time. :)

### DIFF
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -1,0 +1,45 @@
+"""Handles all file writing and post-installation processing."""
+
+from installer._compat.typing import TYPE_CHECKING
+from installer.records import Record
+
+if TYPE_CHECKING:
+    from typing import BinaryIO, Iterable
+
+    from installer._compat.typing import FSPath
+    from installer.utils import Scheme
+
+
+class WheelDestination(object):
+    """Represents the location for wheel installation.
+
+    Subclasses are expected to handle script generation and rewriting of the
+    RECORD file after installation.
+    """
+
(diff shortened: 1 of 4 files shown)


src/installer/destinations.py
```diff
@@ -0,0 +1,45 @@
...
+class WheelDestination(object):
+    """Represents the location for wheel installation.
+
+    Subclasses are expected to handle script generation and rewriting of the
+    RECORD file after installation.
+    """
+
+    def write_file(self, scheme, path, stream):
+        # type: (Scheme, FSPath, BinaryIO) -> Record
+        """TODO: write a good one line description of this function.
```
pradyunsg, line 22: Suggestions welcome!

pradyunsg (commented):
FWIW, I'm not looking for feedback on the abstractions just yet -- that should wait until the core logic is implemented. :)

PR #29, opened 2020-12-18
Add the initial `install` implementation

pradyunsg:


src/installer/utils.py
```diff
@@ -67,3 +73,33 @@ def parse_wheel_filename(filename):
     if not wheel_info:
         raise ValueError("Not a valid wheel filename: {}".format(filename))
     return WheelFilename(*wheel_info.groups())
+
+
+def copyfileobj_with_hashing(
```
pradyunsg, line 78: Any bright ideas for a better name for this function? :3

brettcannon: This looks to have been renamed, so marking as "resolved".

pradyunsg (commented):
Merging since I wanna sort out the documentation deployment, which needs this to be merged.

Happy to get post-merge reviews and to file follow up PRs. :)

PR #30, opened 2020-12-19
record: rename Record to RecordEntry

FFY00:
The Record class only represents on record entry in the RECORD file,
RecordEntry makes that a bit more clear.


pradyunsg (approved):

pradyunsg:
This is based on a chat that @FFY00 and I had; where we agreed that `RecordEntry` is probably a clearer name than `Record` for this.

PR #31, opened 2020-12-19
Add WheelDestination.write_script

pradyunsg:
Shoutout to @FFY00, coz I hadn't thought thoroughly about this. :)

### DIFF
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -7,6 +7,7 @@ if TYPE_CHECKING:
     from typing import BinaryIO, Iterable
 
     from installer._compat.typing import FSPath
+    from installer.scripts import ScriptSection
     from installer.utils import Scheme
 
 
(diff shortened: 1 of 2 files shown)


FFY00:
I was implementing this and it turns out we actually need two methods. One to generate and install entrypoint scripts, and one to install scripts from `{distribution}-{version}.data/scripts`. This is because PEP 427 recommends replacing `#!python` and `#!pythonw` in `*.data/scripts` scripts.

I propose having `write_entrypoint_script` and `write_script`.

Users that want to opt out of the recommended shebang replacement can call `write_file` directly instead of `write_script`.

FFY00:
Similarly, if `write_script` is not implemented, we can fallback to `write_file`.

pradyunsg:
After having a chat w/ @FFY00...

This could be renamed to `write_entrypoint_script`. I'm defering that for now, but I think that might help avoid confusion.

> I was implementing this and it turns out we actually need two methods. One to generate and install entrypoint scripts, and one to install scripts from `{distribution}-{version}.data/scripts`.

This is covered already, since `write_file` would get those files, not `write_script`.

PR #34, opened 2020-12-19
pre-commit: update black to Python 3.9

FFY00:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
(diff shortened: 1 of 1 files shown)


pradyunsg:
https://github.com/pradyunsg/installer/blob/main/noxfile.py#L23

This is intentionally 3.8 — since that’s the same as all the sessions in the noxfile, in case some contributor only has single version of Python in their environment. Bump everything to 3.9 if you’d prefer. :)

pradyunsg:
Oh, and you’ll also need to update GA.

FFY00:
Eh, we can close.

PR #35, opened 2020-12-19
Add initial implementation of `SchemeDictionaryDestination`

FFY00:

### DIFF
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -3,6 +3,7 @@
 import io
 import os.path
 
+from installer._compat import FileExistsError
 from installer._compat.typing import TYPE_CHECKING
 from installer.records import Hash, RecordEntry
 from installer.scripts import Script
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -1,13 +1,18 @@
 """Handles all file writing and post-installation processing."""
 
+import io
+import os.path
+
 from installer._compat.typing import TYPE_CHECKING
-from installer.records import RecordEntry
+from installer.records import Hash, RecordEntry
+from installer.scripts import Script
+from installer.utils import construct_record_file, copyfileobj_with_hashing, fix_shebang
 
 if TYPE_CHECKING:
-    from typing import BinaryIO, Iterable
+    from typing import BinaryIO, Dict, Iterable, Tuple
 
     from installer._compat.typing import FSPath
-    from installer.scripts import ScriptSection
+    from installer.scripts import LauncherKind, ScriptSection
     from installer.utils import Scheme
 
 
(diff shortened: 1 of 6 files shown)


pradyunsg:
https://github.com/pradyunsg/installer/blob/main/noxfile.py#L23

This is intentionally 3.8 — since that’s the same as all the sessions in the noxfile, in case some contributor only has single version of Python in their environment. Bump everything to 3.9 if you’d prefer. :)

src/installer/destinations.py
```diff
@@ -55,3 +61,59 @@ def finalize_installation(self, scheme, record_file_path, records):
...
+        return self._write_file("scripts", name, io.BytesIO(data))
+
+    def finalize_installation(self, scheme, record_file_path, records):
+        # type: (Scheme, FSPath, Iterable[RecordEntry]) -> None
+        """Write the RECORD file and generate the Python cache."""
+        record = RecordEntry("RECORD", None, None)
+        self._write_file(
+            scheme,
+            record_file_path,
+            construct_record_file(list(records) + [record]),
```
pradyunsg, line 117: This BytesIO needs to be closed.

I'd make the helper into a context manager with contextlib. The

tests/test_destinations.py
```diff
@@ -22,3 +27,95 @@ def test_raises_not_implemented_error(self):
                 record_file_path=None,
                 records=None,
             )
+
+
+class TestSchemeDictionaryDestination:
+    @pytest.fixture()
+    def destination(self, tmp_path):
+        schemes = (
```
pradyunsg, line 35: Use SCHEME_NAMES here? Those are taken directly from the PEP.

src/installer/utils.py
```diff
@@ -103,3 +107,49 @@ def copyfileobj_with_hashing(
...
+        return "win-arm64"
+    if "(arm)" in sys.version.lower():
+        return "win-arm"
+    if sys.platform == "win32":
+        return "win-ia32"
+
+    raise NotImplementedError("Unknown launcher kind for this machine")
+
+
+def fix_shebang(data, interpreter):
```
pradyunsg, line 129: I'd call it stream and new_stream -- data implies that it's a bytes/string to me.

pradyunsg: Same note about closing the BytesIO applies.

pradyunsg (commented):
A few notes, but looks great overall!

The linter isn't complaining so I will: please add docstrings to the functions in utils. 🙃

src/installer/destinations.py
pradyunsg: The other option is to clearly state that the end user is responsible for closing this stream, which is also fine by me.

FFY00: Why does it need to be closed by us? I would assume otherwise.

```python
# source
with open('some_file') as f:
    destinatination.write_file('data', 'some_file', f)
```

FFY00: `WheelSource.get_contents` returns an iterable, I would assume it would be implemented like this:

```python
def get_contents(self):
    for file in ...:
        with open(file) as stream:
            yield ..., stream
```

pradyunsg: `construct_record_file`'s return value is not being closed anywhere.

src/installer/utils.py
```diff
@@ -129,7 +129,10 @@ def get_launcher_kind():  # pragma: no cover
 
 def fix_shebang(stream, interpreter):
     # type: (BinaryIO, str) -> BinaryIO
-    """Replace ^#!python shebang in a stream with the correct interpreter."""
+    """Replace ^#!python shebang in a stream with the correct interpreter.
+
+    The original stream should be closed by the caller.
```
pradyunsg, line 134: And the returned stream?

FFY00: Also.

src/installer/destinations.py
FFY00: I fixed that.

src/installer/utils.py
FFY00: I made `fix_shebang` a context manager because it is in a weird position where it may either return the original stream or a new one. The caller doesn't know if they should explicitly close the stream we return or not.
Using a context manager allows us to keep the original stream open and close ours if needed.

src/installer/utils.py
FFY00: See comment above. The returned stream now does not need to be closed, only the original one does.

FFY00:
Friendly ping :blush:

PR #50, opened 2021-05-03
 destinations: SchemeDictionaryDestination._write_file to write_to_fs 

### LINKED ISSUE #49: Make `SchemeDictionaryDestination._write_file` public
Hey, I am implementing a CLI and need to have a `destdir` option that allows me to relocate the install root to an arbitrary path. This is used when building distro packages for example, because we want to put the files in the package directory instead of on the system. This analogous to the `setup.py install --root` option.

Would it be possible to make `SchemeDictionaryDestination._write_file` public so that I can simply overwrite it, instead of having to overwrite `write_file` and `write_script`.
I recommend renaming it to `write_to_fs`, what do you think?

FFY00:
Fixes #49

### DIFF
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -8,18 +8,20 @@ import re
 import sys
 from collections import namedtuple
 from email.parser import FeedParser
+from typing import NewType
 
 from installer._compat import ConfigParser
 from installer._compat.typing import TYPE_CHECKING, Text, cast
 
+Scheme = NewType("Scheme", str)
+
 if TYPE_CHECKING:
     from email.message import Message
-    from typing import BinaryIO, Iterable, Iterator, NewType, Tuple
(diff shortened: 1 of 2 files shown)


FFY00:
While implementing this I realized `installer.utils.Scheme` is not available at runtime, and it is required to instantiate schemes, so I had to remove it from the type guards.

src/installer/utils.py
```diff
@@ -8,18 +8,20 @@
 import sys
 from collections import namedtuple
 from email.parser import FeedParser
+from typing import NewType
```
pradyunsg, line 11: You broke Python 2. :)

pradyunsg: Ah, we're declaring an explicit dependency on this. Nvm me.

PR #63, opened 2021-06-21
Fix entrypoints management

sdispater:
I am currently trying to integrate `installer` into Poetry and things are mostly good but there were issues with how the entrypoints were managed. More specifically:

- The entry points file was not properly detected (`installer` was looking for a `entry-points.txt` file while the actual file is `entry_points.txt`).
- Assertion errors were raised for some entry points (either because they did not match the regexp or were missing the attribute part, which is not required for non-script entry points). This was fixed by ensuring that the `parse_entrypoints()` function only parses script entry points.
- Finally, generated scripts were not executable.

### DIFF
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -190,6 +190,9 @@ def parse_entrypoints(text):
 
     # Borrowed from https://github.com/python/importlib_metadata/blob/v3.4.0/importlib_metadata/__init__.py#L90  # noqa
     for section in config.sections():
+        if section not in ["console_scripts", "gui_scripts"]:
+            continue
+
         for name, value in config.items(section):
             assert isinstance(name, Text)
             match = _ENTRYPOINT_REGEX.match(value)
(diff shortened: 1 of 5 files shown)


uranusjr (approved):

src/installer/utils.py
```diff
@@ -190,6 +190,9 @@ def parse_entrypoints(text):
 
     # Borrowed from https://github.com/python/importlib_metadata/blob/v3.4.0/importlib_metadata/__init__.py#L90  # noqa
```
uranusjr, line 191: Is this comment still up-to-date @pradyunsg? The linked target is entirely different from the code below.

pradyunsg: Yea, this is out of date. Let’s drop it.

@sdispater could you make this change in this PR?

pradyunsg (approved):

src/installer/utils.py
pradyunsg: Actually, nvm. Let’s do that in a follow up.

sdispater: @pradyunsg I did it anyway :-)

frostming:
Hi, when will this fix be released? I am planning to adopt `installer`  in PDM.

pradyunsg:
Ah buggery. I forgot about that.

I’ll do if over the coming weekend but if I don’t, please file a new issue.

PR #72, opened 2021-09-13
Fix for wheels compressed with zip instead of python wheels package

### LINKED ISSUE #71: Zipinfo Directories
Whilst trying to solve a problem with installing torch using PDM I came across #69. After reading the code I realized that the change provided makes the error message more verbose but doesn't solve the issue for the torch package. 

The problem seems to be that some whl files, when looping over `self._zipfile.infolist()` include the directories as well as the files which is what causes that function to fail because the RECORD file doesn't have information about directories. eg:

Torch (Directories as part of the list):
```
torch/
torch/_deploy.py
torch/functional.py
torch/onnx/
torch/onnx/symbolic_opset9.py
torch/onnx/symbolic_caffe2.py
torch/onnx/symbolic_helper.py
torch/onnx/symbolic_opset12.py
torch/onnx/symbolic_registry.py
```

Colorama (Files only):
```
colorama/__init__.py
colorama/ansi.py
colorama/ansitowin32.py
colorama/initialise.py
colorama/win32.py
colorama/winterm.py
colorama-0.4.4.dist-info/LICENSE.txt
colorama-0.4.4.dist-info/METADATA
colorama-0.4.4.dist-info/WHEEL
colorama-0.4.4.dist-info/top_level.txt
colorama-0.4.4.dist-info/RECORD
```

Omegaice:
When using zip to compress a wheel package it adds entries for the directories themselves which breaks this code because the code expects every entry in the zip file to have an entry in the RECORD file.

Fixes #71


FFY00 (approved):

src/installer/sources.py
```diff
@@ -162,12 +162,15 @@ def get_contents(self):
...
+                assert (
+                    record is not None
+                ), "In {}, {} is not mentioned in RECORD".format(
+                    self._zipfile.filename,
+                    item.filename,
+                )  # should not happen for valid wheels
+
+                with self._zipfile.open(item) as stream:
+                    stream_casted = cast("BinaryIO", stream)
+                    yield record, stream_casted
```
pradyunsg, line 176:
```suggestion
            if item.filename[-1] == "/":  # looks like a directory
                continue

            record = record_mapping.pop(item.filename, None)
            assert record is not None, "In {}, {} is not mentioned in RECORD".format(
                self._zipfile.filename,
                item.filename,
            )  # should not happen for valid wheels

            with self._zipfile.open(item) as stream:
                stream_casted = cast("BinaryIO", stream)
                yield record, stream_casted
```

pradyunsg, line 176:
```suggestion
            if item.filename[-1] == "/":  # looks like a directory
                continue

            record = record_mapping.pop(item.filename, None)
            assert record is not None, "In {}, {} is not mentioned in RECORD".format(
                self._zipfile.filename,
                item.filename,
            )  # should not happen for valid wheels

            with self._zipfile.open(item) as stream:
                stream_casted = cast("BinaryIO", stream)
                yield record, stream_casted
```

pradyunsg:
Ah fun. Coverage is able to realise now that we never hit this branch in the tests.

Would you be willing to add a test for this @Omegaice? If not, I’ll be happy to pick this up instead.

Omegaice:
I just pushed a commit which should have a valid test. It passes locally but I didn't run the full gamut of tests as I don't have all the python versions installed and pre-commit had some issues with something to do with GPG. 

I had to edit the test in `test_provides_correct_contents` to filter directories so that `assert got_files == files` didn't fail and I also had to edit `dist_info_filenames` to also filter out directories to make it's test pass correctly.

src/installer/sources.py
```diff
@@ -139,6 +139,7 @@ def dist_info_filenames(self):
         return [
             name[len(base) + 1 :]
             for name in self._zipfile.namelist()
+            if name[-1] != "/"
```
uranusjr, line 142: I wonder if we need to account for an empty string here. (easiest solution would be `name[-1:] != "/"`)

tests/test_sources.py
```diff
@@ -13,6 +13,7 @@
 def fancy_wheel(tmp_path):
     path = tmp_path / "fancy-1.0.0-py2.py3-none-any.whl"
     files = {
+        "fancy/": b"""""",
```
uranusjr, line 16: If feel these directory entries should use `None` as values instead (and change the check below). Hard-coding the `/` logic in tests does not feel right to me.

src/installer/sources.py
Omegaice: I don't think this should be necessary because it looks at entries inside of the ZIP file but I will add it for extra safety.

tests/test_sources.py
Omegaice: It seems like the only way to get ZipFile to add directory entries is to write an empty item that ends with "/". Setting it to `"fancy/": None,` breaks when it tries to write the data to the entry.

I do agree that having the test effectively running the same code as the function we are testing is not the best test implementation. It might be that pulling out the functionality of getting only the files from the zip would be a cleaner way of handling it.

Omegaice:
I also pushed to another branch some exploration with restructuring to reduce the number of times the directory check is handled. Let me know if you feel it is relevant and I can merge it into this branch. https://github.com/Omegaice/installer/commit/9f4a1867350fe4c5f7e845f075efbd9ab04a0611

tests/test_sources.py
uranusjr: Quite sad, but I guess we can only use what's available to us.

pradyunsg:
> Let me know if you feel it is relevant

It's adding additional public API methods on the `WheelFile` class, which I strongly dislike. Even if those were private though, I don't think there's much value to removing the duplication of a signle conditional that shows up only twice + once in the tests.

Omegaice:
That was the reason I ended up pushing it to a different branch. Is there anything else that would benefit this PR?

pradyunsg:
Nope, we're good here!

Thanks a ton for contributing this fix, being patient about this and working collaboratively to iterate on this PR! :)

whiskeyriver:
Hit this with greenlet/sqlalchemy as well.

PR #75, opened 2021-09-25
core: return written records in install

FFY00:


FFY00:
This is needed for the bytecode generation step afterwards. This commit is/was part of #66 but I have created a separate PR to speed up review. This is the only thing preventing me from providing a CLI in a 3rd party package.

src/installer/_core.py
```diff
@@ -69,9 +72,11 @@ def _determine_scheme(path, source, root_scheme):
 
 
 def install(source, destination, additional_metadata):
-    # type: (WheelSource, WheelDestination, Dict[str, Binary]) -> None
+    # type: (WheelSource, WheelDestination, Dict[str, Binary]) -> Mapping[str, Iterable[RecordEntry]]
```
pradyunsg, line 75: I'm a strong -1 on this.

I'd much rather change the interface for the `WheelDestination.finalise_installation` than add a return type to this function. The literal job of `WheelDestination.finalise_installation` is to do things like this.

pradyunsg (changes requested):

FFY00:
@pradyunsg friendly ping.

FFY00:
Thank you, I will have a look tomorrow.

pradyunsg:
> Sorry for the delay.

No worries! ^>^

Here you go: https://pypi.org/project/installer/0.4.0/

src/installer/utils.py
```diff
@@ -169,13 +169,13 @@ def fix_shebang(stream, interpreter):
...
 def construct_record_file(records):
-    # type: (Iterable[RecordEntry]) -> BinaryIO
+    # type: (Iterable[Tuple[Scheme, RecordEntry]]) -> BinaryIO
     """Construct a RECORD file given some records.
 
     The original stream should be closed by the caller.
     """
     stream = io.BytesIO()
-    for record in records:
+    for scheme, record in records:
```
frostming, line 178: `scheme` isn't used.

pradyunsg: Not here, but it's used in other places that this list of records comes from.

Issue #58, opened 2021-05-15
Enable installation under a specified prefix


pradyunsg:
There are requests for baking in support for https://gnu.org/prep/standards/html_node/DESTDIR.html in this library.

> This is a basic feature in any kind of installer, and without it packagers cannot use this tool.

_Originally posted by @FFY00 in https://github.com/pradyunsg/installer/issues/52#issuecomment-835831533_

> Yeah, use-sysconfig-but-chroot-first is plain necessary for this common use case.

_Originally posted by @flying-sheep in https://github.com/pradyunsg/installer/issues/52#issuecomment-838092371_

pradyunsg:
I don't think destdir support needs to be baked into any Destination object, or need special casing in this library. The caller can do the prefixing themselves.

```py
def prefix_with_destdir(path: str) -> str:
    return os.path.join(my_destdir, path.lstrip("/"))


schemes = {key: prefix_with_destdir(value) for key, value in sysconfig.get_config_vars()}
interpreter = sys.executable
my_destdir_destination = SchemeDictDestination(schemes, interpreter, "posix")
```

_Originally posted by @pradyunsg in https://github.com/pradyunsg/installer/issues/52#issuecomment-841639249_

FFY00:
This is the incorrect fix, I don't think anything uses full paths from sysconfig right now, but they could in the future. The project would not be ready for such cases, in fact, the implementation would be incorrect. This really is a feature fro the destionation, I just want to place the file in a custom path during the copying/install process, everything else should be the same.

If I implement it in a separate library/CLI it would be via `write_to_fs`, not by changing the sysconfig paths. But that is not the issue, the issue is that it would be extermetlly helpful if this was supported by the CLI of this project.
If it isn't, this project no longer becomes viable for Linux distros. We have to write a separate project for that which will just make the bootstrapping process even more difficult. Given how easy it is to implement, and how common of the feature it is in any kind of installer, I don't understand why it can't be supported by this project. I am even willing to contribute all the code and tests, which isn't that much anyway, and commit to maintainership of it.

pradyunsg:
There's 3 things here:

- the example isn't demonstrating "use sysconfig" but rather that you don't need dedicated custom code in SchemeDictionaryDestination to do the prefix based stuff.

- if you want to inherit and override, you're free to. I don't think that gets you anything, if all you're doing is putting a prefix to a path that you provided yourself.

[...]

PR #89, opened 2021-12-11
Allow changing the installation root.

jameshilliard:
Should provide distutils/setuptools style `--root` change functionality handy for cross compilation.

### DIFF
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -2,6 +2,7 @@
 
 import io
 import os
+import sysconfig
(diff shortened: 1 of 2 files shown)


FFY00:
The logic here is a bit fragile and this issue has already been extensively discussed, see #58. #66 implements it as a `destdir` option, aligning the naming with common practices and other installation systems (autotools, cmake, meson, etc.).

jameshilliard:
> #66 implements it as a `destdir` option

Oh, had missed that for some reason.

PR #90, opened 2021-12-11
Preserve the executable bit from wheel contents

### LINKED ISSUE #68: The executable bit is lost for scripts in data directory
`installer` seems only to copy streams and won't copy the executable bit when installing

pradyunsg:
This is a property of individual files in the wheel, and thus needs to be parsed
by the `WheelSource`, and needs special handling in `WheelDestination`. It also
needs to be propagated through the entire pipeline, since the information needs
to be exchanged between the two components.

Closes #68

### DIFF
--- a/src/installer/sources.py
+++ b/src/installer/sources.py
@@ -2,6 +2,7 @@
 
 import os
 import posixpath
+import stat
 import zipfile
 from contextlib import contextmanager
 from typing import BinaryIO, Iterator, List, Tuple, cast
@@ -9,7 +10,7 @@ from typing import BinaryIO, Iterator, List, Tuple, cast
 import installer.records
 import installer.utils
 
-WheelContentElement = Tuple[Tuple[str, str, str], BinaryIO]
+WheelContentElement = Tuple[Tuple[str, str, str], BinaryIO, bool]
 
 
 __all__ = ["WheelSource", "WheelFile"]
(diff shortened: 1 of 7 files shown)


frostming:
May I know when will this fix be released?

src/installer/sources.py
```diff
@@ -2,14 +2,15 @@
...
+import stat
 import zipfile
 from contextlib import contextmanager
 from typing import BinaryIO, Iterator, List, Tuple, cast
 
 import installer.records
 import installer.utils
 
-WheelContentElement = Tuple[Tuple[str, str, str], BinaryIO]
+WheelContentElement = Tuple[Tuple[str, str, str], BinaryIO, bool]
```
takluyver, line 13: YMMV, but when I add the third field, I tend to start wondering if it would be clearer with a dataclass rather than a tuple.

pradyunsg: Yea, I had the same thought too. :)

However, I want this to be decoupled such that the source can be implemented without a dependency on this package. I'm basically thinking that `wheel` would end up implementing this interface, in their public API, and then we'd have a nice amount of synergy.

pradyunsg:
> May I know when will this fix be released?

Sometime next month. Most of my OSS time this month is gonna go into pip's release.

PR #94, opened 2022-01-05
CLI to install to Python running installer

### LINKED ISSUE #93: Add support for pycompile (as an opt-out)
Based off https://github.com/pradyunsg/installer/pull/66#discussion_r763810443

takluyver:
This starts from #66, and moves the added functionality for destdir & bytecode compilation into the Python API (`SchemeDictionaryDestination`), so the CLI is a thinner wrapper over the Python API. It also drops the compatibility checks, so `installer` once again has no dependencies.

Alternative to #92, for consideration.
Closes #93

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -4,11 +4,8 @@ import argparse
 import compileall
 import distutils.dist
 import pathlib
-import platform
 import sys
 import sysconfig
-import warnings
-from email.message import Message
 from typing import TYPE_CHECKING, Collection, Dict, Iterable, Optional, Sequence, Tuple
 
 import installer
@@ -20,10 +17,6 @@ from installer.utils import Scheme
 
 if TYPE_CHECKING:
     from installer.scripts import LauncherKind
-
-
-class InstallerCompatibilityError(Exception):
-    """Error raised when the install target is not compatible with the environment."""
 
 
 class _MainDestination(installer.destinations.SchemeDictionaryDestination):
@@ -102,12 +95,6 @@ def main_parser() -> argparse.ArgumentParser:
         default=(0, 1),
         help="generate bytecode for the specified optimization level(s) (default=0, 1)",
     )
-    parser.add_argument(
-        "--skip-dependency-check",
-        "-s",
-        action="store_true",
-        help="don't check if the wheel dependencies are met",
-    )
     return parser
 
 
@@ -130,52 +117,6 @@ def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
     return scheme_dict
 
 
-def check_python_version(metadata: Message) -> None:
-    """Check if the project support the current interpreter."""
-    try:
-        import packaging.specifiers
-    except ImportError:
-        warnings.warn(
-            "'packaging' module not available, "
-            "skipping python version compatibility check"
-        )
-        return
-
-    requirement = metadata["Requires-Python"]
-    if not requirement:
-        return
-
-    versions = requirement.split(",")
-    for version in versions:
-        specifier = packaging.specifiers.Specifier(version)
-        if platform.python_version() not in specifier:
-            raise InstallerCompatibilityError(
-                "Incompatible python version, needed: {}".format(version)
-            )
-
-
-def check_dependencies(metadata: Message) -> None:
-    """Check if the project dependencies are met."""
-    try:
-        import build
-        import packaging  # noqa: F401
-    except ModuleNotFoundError as e:
-        warnings.warn(f"'{e.name}' module not available, skipping dependency check")
-        return
-
-    missing = {
-        unmet
-        for requirement in metadata.get_all("Requires-Dist") or []
-        for unmet_list in build.check_dependency(requirement)
-        for unmet in unmet_list
-    }
-    if missing:
-        missing_list = ", ".join(missing)
-        raise InstallerCompatibilityError(
-            "Missing requirements: {}".format(missing_list)
-        )
-
-
 def main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
     """Process arguments and perform the install."""
     parser = main_parser()
@@ -184,13 +125,6 @@ def main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
     args = parser.parse_args(cli_args)
 
     with installer.sources.WheelFile.open(args.wheel) as source:
-        # compability checks
-        metadata_contents = source.read_dist_info("METADATA")
-        metadata = installer.utils.parse_metadata_file(metadata_contents)
-        check_python_version(metadata)
-        if not args.skip_dependency_check:
-            check_dependencies(metadata)
-
         destination = _MainDestination(
             get_scheme_dict(source.distribution),
             sys.executable,
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -82,5 +82,5 @@ def main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
         installer.install(source, destination, {})
 
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: no cover
     main(sys.argv[1:], "python -m installer")
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -46,11 +46,8 @@ def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
 
     # calculate 'headers' path, not currently in sysconfig - see
     # https://bugs.python.org/issue44445. This is based on what distutils does.
-    # For new wheels, the name we get from the wheel filename should already
-    # be normalised (based on PEP 503, but with _ instead of -), but this is
-    # not the case for many existing wheels, so normalise it here.
-    normed_name = re.sub(r"[-_.]+", "_", distribution_name).lower()
-    scheme_dict["headers"] = os.path.join(scheme_dict["include"], normed_name)
+    # TODO: figure out original vs normalised names
+    scheme_dict["headers"] = os.path.join(scheme_dict["include"], distribution_name)
 
     return scheme_dict
 
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -2,6 +2,7 @@
 
 import argparse
 import os.path
+import re
 import sys
 import sysconfig
 from typing import Dict, Optional, Sequence
@@ -43,9 +44,13 @@ def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
     """Calculate the scheme dictionary for the current Python environment."""
     scheme_dict = sysconfig.get_paths()
 
-    # calculate 'headers' path, not currently in sysconfig -
-    # see https://bugs.python.org/issue44445. This copies what distutils does.
-    scheme_dict["headers"] = os.path.join(scheme_dict["include"], distribution_name)
+    # calculate 'headers' path, not currently in sysconfig - see
+    # https://bugs.python.org/issue44445. This is based on what distutils does.
+    # For new wheels, the name we get from the wheel filename should already
+    # be normalised (based on PEP 503, but with _ instead of -), but this is
+    # not the case for many existing wheels, so normalise it here.
+    normed_name = re.sub(r"[-_.]+", "_", distribution_name).lower()
+    scheme_dict["headers"] = os.path.join(scheme_dict["include"], normed_name)
 
     return scheme_dict
 
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -157,9 +157,8 @@ def check_python_version(metadata: Message) -> None:
 def check_dependencies(metadata: Message) -> None:
     """Check if the project dependencies are met."""
     try:
-        import packaging  # noqa: F401
-
         import build
+        import packaging  # noqa: F401
     except ModuleNotFoundError as e:
         warnings.warn(f"'{e.name}' module not available, skipping dependency check")
         return
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -2,7 +2,6 @@
 
 import argparse
 import os.path
-import re
 import sys
 import sysconfig
 from typing import Dict, Optional, Sequence
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -24,13 +24,17 @@ def main_parser() -> argparse.ArgumentParser:
         help="destination directory (prefix to prepend to each file)",
     )
     parser.add_argument(
-        "--optimize",
-        "-o",
-        nargs="*",
+        "--compile-bytecode",
+        action="append",
         metavar="level",
         type=int,
-        default=(0, 1),
+        choices=[0, 1, 2],
         help="generate bytecode for the specified optimization level(s) (default=0, 1)",
+    )
+    parser.add_argument(
+        "--no-compile-bytecode",
+        action="store_true",
+        help="don't generate bytecode for installed modules",
     )
     return parser
 
@@ -61,12 +65,18 @@ def main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
         parser.prog = program
     args = parser.parse_args(cli_args)
 
+    bytecode_levels = args.compile_bytecode
+    if args.no_compile_bytecode:
+        bytecode_levels = []
+    elif not bytecode_levels:
+        bytecode_levels = [0, 1]
+
     with installer.sources.WheelFile.open(args.wheel) as source:
         destination = installer.destinations.SchemeDictionaryDestination(
             get_scheme_dict(source.distribution),
             sys.executable,
             installer.utils.get_launcher_kind(),
-            bytecode_optimization_levels=args.optimize,
+            bytecode_optimization_levels=bytecode_levels,
             destdir=args.destdir,
         )
         installer.install(source, destination, {})
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -21,7 +21,6 @@ def main_parser() -> argparse.ArgumentParser:
         "-d",
         metavar="/",
         type=str,
-        default="/",
         help="destination directory (prefix to prepend to each file)",
     )
     parser.add_argument(
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -1,0 +1,211 @@
+"""Installer CLI."""
+
+import argparse
+import compileall
+import distutils.dist
+import pathlib
+import platform
+import sys
+import sysconfig
+import warnings
+from email.message import Message
+from typing import TYPE_CHECKING, Collection, Dict, Iterable, Optional, Sequence, Tuple
+
+import installer
+import installer.destinations
+import installer.sources
+import installer.utils
+from installer.records import RecordEntry
+from installer.utils import Scheme
+
+if TYPE_CHECKING:
+    from installer.scripts import LauncherKind
+
+
+class InstallerCompatibilityError(Exception):
+    """Error raised when the install target is not compatible with the environment."""
+
+
+class _MainDestination(installer.destinations.SchemeDictionaryDestination):
+    destdir: Optional[pathlib.Path]
+
+    def __init__(
+        self,
+        scheme_dict: Dict[str, str],
+        interpreter: str,
+        script_kind: "LauncherKind",
+        hash_algorithm: str = "sha256",
+        optimization_levels: Collection[int] = (0, 1),
+        destdir: Optional[str] = None,
+    ) -> None:
+        if destdir:
+            self.destdir = pathlib.Path(destdir).absolute()
+            self.destdir.mkdir(exist_ok=True, parents=True)
+            scheme_dict = {
+                name: self._destdir_path(value) for name, value in scheme_dict.items()
+            }
+        else:
+            self.destdir = None
+        super().__init__(scheme_dict, interpreter, script_kind, hash_algorithm)
+        self.optimization_levels = optimization_levels
+
+    def _destdir_path(self, file: str) -> str:
+        assert self.destdir
+        file_path = pathlib.Path(file)
+        rel_path = file_path.relative_to(file_path.anchor)
+        return str(self.destdir.joinpath(*rel_path.parts))
+
+    def _compile_record(self, scheme: Scheme, record: RecordEntry) -> None:
+        if scheme not in ("purelib", "platlib"):
+            return
+        for level in self.optimization_levels:
+            target_path = pathlib.Path(self.scheme_dict[scheme], record.path)
+            if sys.version_info < (3, 9):
+                compileall.compile_file(target_path, optimize=level)
+            else:
+                compileall.compile_file(
+                    target_path,
+                    optimize=level,
+                    stripdir=str(self.destdir),
+                )
+
+    def finalize_installation(
+        self,
+        scheme: Scheme,
+        record_file_path: str,
+        records: Iterable[Tuple[Scheme, RecordEntry]],
+    ) -> None:
+        record_list = list(records)
+        super().finalize_installation(scheme, record_file_path, record_list)
+        for scheme, record in record_list:
+            self._compile_record(scheme, record)
+
+
+def main_parser() -> argparse.ArgumentParser:
+    """Construct the main parser."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("wheel", type=str, help="wheel file to install")
+    parser.add_argument(
+        "--destdir",
+        "-d",
+        metavar="/",
+        type=str,
+        default="/",
+        help="destination directory (prefix to prepend to each file)",
+    )
+    parser.add_argument(
+        "--optimize",
+        "-o",
+        nargs="*",
+        metavar="level",
+        type=int,
+        default=(0, 1),
+        help="generate bytecode for the specified optimization level(s) (default=0, 1)",
+    )
+    parser.add_argument(
+        "--skip-dependency-check",
+        "-s",
+        action="store_true",
+        help="don't check if the wheel dependencies are met",
+    )
+    return parser
+
+
+def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
+    """Calculate the scheme disctionary for the current Python environment."""
+    scheme_dict = sysconfig.get_paths()
+
+    # calculate 'headers' path, sysconfig does not have an equivalent
+    # see https://bugs.python.org/issue44445
+    dist_dict = {
+        "name": distribution_name,
+    }
+    distribution = distutils.dist.Distribution(dist_dict)
+    install_cmd = distribution.get_command_obj("install")
+    assert install_cmd
+    install_cmd.finalize_options()
+    # install_cmd.install_headers is not type hinted
+    scheme_dict["headers"] = install_cmd.install_headers  # type: ignore
+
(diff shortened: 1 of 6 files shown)


src/installer/__main__.py
```diff
@@ -0,0 +1,82 @@
...
+        help="destination directory (prefix to prepend to each file)",
+    )
+    parser.add_argument(
+        "--optimize",
+        "-o",
+        nargs="*",
+        metavar="level",
+        type=int,
+        default=(0, 1),
+        help="generate bytecode for the specified optimization level(s) (default=0, 1)",
```
jameshilliard, line 34: is this able to be turned off entirely if desired?

pradyunsg:
Two major concerns:

- Making the compilation opt-out on the CLI.
- Ensuring that the destdir=/ works on Windows. It might be a good idea to have it default to None instead.

takluyver:
I agree, I don't think `/` would do the right thing on Windows, so I've made `None` the default.

For bytecode compilation, you can *technically* tell it not to by passing `--optimize` with no levels. Obviously that's a pretty back-to-front interface, though.

Do you want bytecode generation to be on by default, with separate arguments to disable it (`--no-compile-bytecode`?) and to control the optimisation levels? Or off by default, and you specify e.g. `--compile-bytecode 0 1` to enable it? Maybe allowing a plain `--compile-bytecode` with no value to enable it for level 0?

pradyunsg:
How about:

- no-args: compile with level 0 and 1.
- `--no-compile-bytecode`: don't compile.
- `--compile-bytecode 0`: compile with level 0 only.
-  `--compile-bytecode 0 --compile-bytecode 1 --compile-bytecode 2`: compile with level 0 and level 1 and level 2.

src/installer/__main__.py
pradyunsg: Should be. That's being discussed further down in the thread.

takluyver:
I think that should implement what you suggested. I still need to get to writing some tests.

src/installer/__main__.py
```diff
@@ -0,0 +1,86 @@
...
+    parser.add_argument(
+        "--no-compile-bytecode",
+        action="store_true",
+        help="don't generate bytecode for installed modules",
+    )
+    return parser
+
+
+def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
+    """Calculate the scheme disctionary for the current Python environment."""
```
mgorny, line 43:
```suggestion
    """Calculate the scheme dictionary for the current Python environment."""
```

src/installer/__main__.py
```diff
@@ -0,0 +1,86 @@
...
+    scheme_dict = sysconfig.get_paths()
+
+    # calculate 'headers' path, sysconfig does not have an equivalent
+    # see https://bugs.python.org/issue44445
+    dist_dict = {
+        "name": distribution_name,
+    }
+    distribution = distutils.dist.Distribution(dist_dict)
+    install_cmd = distribution.get_command_obj("install")
```
mgorny, line 52: FYI, this doesn't seem future-proof — between distutils being deprecated and setuptools deprecating install command.

src/installer/__main__.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def main_parser() -> argparse.ArgumentParser:
+    """Construct the main parser."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("wheel", type=str, help="wheel file to install")
+    parser.add_argument(
+        "--destdir",
+        "-d",
+        metavar="/",
```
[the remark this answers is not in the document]
takluyver: Yup, how about `path`?

src/installer/__main__.py
takluyver: It's certainly not, but I don't think there's any alternative way to get the `headers` destination directory described in the wheel spec. There's a discussion about this here: https://discuss.python.org/t/clarification-on-a-wheels-header-data/9305 and a Python bug open: https://bugs.python.org/issue44445

I've so far not found any wheel that actually uses the headers directory - if that's generally true, then maybe the resolution for the future will be just to remove the `header` directory from the spec, so we can get all the paths from `sysconfig`. Otherwise, it will have to be added to the paths `sysconfig` knows about.

src/installer/__main__.py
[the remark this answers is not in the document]
layday: Sounds good.

src/installer/__main__.py
layday: `greenlet` is one that I use that has a headers folder.

takluyver: Aha, so it is used...

layday: FWIW, pip uses distutils for system installs under Python < 3.10, sysconfig in 3.10 and above, and it synthesises an include path for virtual envs under `sys.prefix`.

pradyunsg (commented):
Apparently I have review comments that I haven't posted?

src/installer/destinations.py
```diff
@@ -93,6 +105,8 @@ def __init__(
         interpreter: str,
         script_kind: "LauncherKind",
         hash_algorithm: str = "sha256",
+        bytecode_optimization_levels: Collection[int] = (0, 1),
```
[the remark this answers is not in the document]
takluyver: -1 means it depends on the running interpreter, if I've followed the Python docs correctly? I think for the Python API, we want to avoid any behaviour that implicitly depends on the interpreter running it. Then again, the bytecode generated is already specific to the minor version of Python we're running on, and I don't think there's any good way to generate bytecode for a different version of Python.

Maybe the default for the Python API should be `()`, i.e. no bytecode compilation by default?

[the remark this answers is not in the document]
pradyunsg: 🤷🏽 

Let's do either `()` or `(0, 1, 2)`. I'll let you pick. :)

src/installer/__main__.py
takluyver: Ugh. So if I'm reading pip's code right, the decision of whether to use paths from sysconfig or distutils is [this code](https://github.com/pypa/pip/blob/7f8a6844037fb7255cfd0d34ff8e8cf44f2598d4/src/pip/_internal/locations/__init__.py#L228-L246), and the calculation of a header scheme directory from sysconfig paths is [this code](https://github.com/pypa/pip/blob/7f8a6844037fb7255cfd0d34ff8e8cf44f2598d4/src/pip/_internal/locations/_sysconfig.py#L173-L191).

I don't think `installer`, as a low-level tool, should be detecting 'am I running in a virtualenv' - encoding high-level concepts in a low-level package like that feels wrong. It also means things behave differently in environments made with e.g. conda or pyenv, which I think goes against user expectations.

On my system, for a standard Posix prefix install of a distribution called `foobar`:

- distutils gives `{prefix}/include/python3.10/foobar` (both with and without a venv)
- pip's sysconfig logic outwith a virtualenv would give `{prefix}/include/python3.10/foobar`
- pip's sysconfig logic inside a virtualenv would give `{prefix}/include/site/python3.10/foobar`

The code in pip explicitly notes that it's somewhat arbitrary and only done for backwards compatibility. For a specific detail like this, I don't think installer should be bound to duplicate pip's behaviour exactly, so I'd suggest we ignore virtualenvs and always use `os.path.join(include_path, distribution_name)` as the header target. I'd guess that's also what an old-school `setup.py install` would do, though I haven't checked that.

src/installer/destinations.py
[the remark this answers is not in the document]
takluyver: I've gone for `()` as the default. At the Python API level, I think people should have to opt in to anything that's specific to the Python version running `installer`.

src/installer/__main__.py
```diff
@@ -0,0 +1,78 @@
...
+    return parser
+
+
+def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
+    """Calculate the scheme dictionary for the current Python environment."""
+    scheme_dict = sysconfig.get_paths()
+
+    # calculate 'headers' path, not currently in sysconfig -
+    # see https://bugs.python.org/issue44445. This copies what distutils does.
+    scheme_dict["headers"] = os.path.join(scheme_dict["include"], distribution_name)
```
layday: This doesn't actually copy what distutils does because (a) `include` is always on the `sys.base_prefix`, meaning that  installer will attempt to install headers under the base prefix, and (b) when the root is "platlib", the installation will be performed under "purelib".

src/installer/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+def get_scheme_dict(distribution_name: str) -> Dict[str, str]:
+    """Calculate the scheme dictionary for the current Python environment."""
+    scheme_dict = sysconfig.get_paths()
+
+    # calculate 'headers' path, not currently in sysconfig - see
+    # https://bugs.python.org/issue44445. This is based on what distutils does.
+    # For new wheels, the name we get from the wheel filename should already
+    # be normalised (based on PEP 503, but with _ instead of -), but this is
+    # not the case for many existing wheels, so normalise it here.
+    normed_name = re.sub(r"[-_.]+", "_", distribution_name).lower()
```
layday, line 52: This shouldn't be necessary, right? The distribution name is extracted from the filename of the wheel.

src/installer/__main__.py
takluyver: I've implemented the what distutils appears to do (make a subdirectory with the distribution name under `include`), but normalising the name like we've agreed to do for other places a distribution name is used to make a file or folder name (wheel filenames, `.dist-info` folders).

@pradyunsg I suspect there's more work to be done on where `installer` uses original vs normalised distribution names. At the moment, the distribution name is parsed out of the wheel filename, but this can either be partially or fully normalised. If we want the original name for anything, that needs to be read from the metadata inside the wheel.

src/installer/__main__.py
layday: Sorry, I totally missed the comment!

takluyver: We've only just agreed that distribution names in wheel filenames should be normalised like this, and it's not yet implemented in wheel. So we have to handle wheel filenames where it's not normalised.

Of course, we could use the original distribution name instead of the normalised one. But I think that requires more changes, because that has to be read from the metadata in the wheel.

layday: Thanks. For anybody else who wasn't aware, the discussion is at https://discuss.python.org/t/revisiting-distribution-name-normalization/12348/15.

layday: (For some reason I can't collapse my own review comments, feel free to.)

mgorny:
I've just tried it and it doesn't seem to work correctly. I've tried:

```
python -m installer -d /tmp/z ~/git/portage/dist/portage-3.0.30-cp310-cp310-linux_x86_64.whl
```

and it installed /etc and /usr into /tmp/z/usr/lib/python3.10/site-packages/ rather than /tmp/z.

src/installer/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+        "-d",
+        metavar="path",
+        type=str,
+        help="destination directory (prefix to prepend to each file)",
+    )
+    parser.add_argument(
+        "--compile-bytecode",
+        action="append",
+        metavar="level",
+        type=int,
```
mgorny, line 31: Could you make it accept a comma-separated list? Right now the help text suggests multiple values can be specified like that, and passing `--compile-bytecode 0 --compile-bytecode 1 --compile-bytecode 2` is a bit long.

src/installer/__main__.py
takluyver: I see what you mean about a) - distutils replaces `installed_base` (= `sys.base_prefix`) with `base` (= `sys.prefix`). We can do the equivalent easily enough, I think.

I'm not sure what you mean about b) though. I don't think that root-is-purelib setting should make any difference to where the headers are installed, and I haven't spotted anything in distutils referring to another destination for them (though I could easily have overlooked it, the logic is fairly spread out).

src/installer/destinations.py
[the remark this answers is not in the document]
pradyunsg: Sounds good!

takluyver:
@mgorny I've just tried it with a portage wheel from PyPI (`py3-none-any`), and I get paths like `/tmp/z/usr/share/man/man1/emerge.1` and `/tmp/z/usr/etc/dispatch-conf.conf`, which I believe are correct based on the contents of the wheel.

If it's different on your system, can you show the result of `sysconfig.get_paths()` ?

src/installer/__main__.py
pradyunsg: Let's leave this normalisation code out of the CLI, and file a follow up issue for API to address it in a blanket move.

takluyver: OK, done.

takluyver: Opened #97 for the name question.

src/installer/destinations.py
```diff
@@ -176,20 +208,32 @@ def write_script(
...
+            path = self._destdir_path(Scheme("scripts"), script_name)
             mode = os.stat(path).st_mode
             mode |= (mode & 0o444) >> 2
             os.chmod(path, mode)
 
             return entry
 
+    def _compile_bytecode(self, scheme: Scheme, record: RecordEntry) -> None:
+        """Compile bytecode for a single .py file."""
+        if scheme not in ("purelib", "platlib"):
```
pradyunsg, line 220: For a follow up: we should consider what the right behaviour for `scripts` would be.

src/installer/__main__.py
layday: I assume that if the root is platlib then the headers should be installed in platinclude, i.e. under `sys.(base_)exec_prefix`.

src/installer/destinations.py
takluyver: AFAIK, Python doesn't even look for cached bytecode for a script it's running, only on imports. Of course, you could have a script that is also importable, but I wouldn't spend too much time thinking about that - the lib directories are the obvious place to put code to be imported.

src/installer/__main__.py
layday: distutils doesn't do this, never mind.

takluyver: I don't see distutils implementing that, and it doesn't seem right to me. The wheel contents are 5 separate components to be installed - purelib and platlib are separate, but there's just one `headers` directory, no `platheaders`. As a convenience, the root of the wheel is considered either platlib or purelib, but I don't think that should affect where any of the other components go.

src/installer/destinations.py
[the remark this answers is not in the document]
eli-schwartz: If the paths to install to are specific to the current interpreter, then it seems reasonable for the bytecode to also be.

If the paths to install are specific to a defined `--interpreter` then that can be used to run `-m compileall`.

src/installer/__main__.py
takluyver: OK, great. I'll deal with prefix vs base_prefix at some point later.

takluyver: The latest change - 275e73daf89de4fde54fcfc2ec187ecbdda22cd4 - should use prefix in place of base_prefix so headers can be installed in a virtualenv.

mgorny:
> @mgorny I've just tried it with a portage wheel from PyPI (`py3-none-any`), and I get paths like `/tmp/z/usr/share/man/man1/emerge.1` and `/tmp/z/usr/etc/dispatch-conf.conf`, which I believe are correct based on the contents of the wheel.

Hmm, interesting. I get the "correct" result from the PyPI wheel as well. Hmm, in fact I'm not really sure where that local wheel came from, maybe I've had a stray file from some experiment. Sorry for the noise.

src/installer/__main__.py
FFY00: Perhaps just add a short name, `-O` (eg. `-O0 -O1`).

src/installer/destinations.py
pradyunsg: Ah, no worries then. :)

src/installer/__main__.py
pradyunsg: Let's leave this as-is for now -- we can always add a shorter form later. :)

src/installer/__main__.py
```diff
@@ -0,0 +1,84 @@
...
+    # https://bugs.python.org/issue44445. This is based on what distutils does.
+    # TODO: figure out original vs normalised distribution names
+    scheme_dict["headers"] = os.path.join(
+        sysconfig.get_path(
+            "include", vars={"installed_base": sysconfig.get_config_var("base")}
+        ),
+        distribution_name,
+    )
+
+    return scheme_dict
```
pradyunsg, line 56: That is... assuming folks are OK with this. :)

pradyunsg (approved):
LGTM!

As a final non-blocking nit-pick: I'd like to rename a private method -- `_destdir_path(...)` to `_path_within_destdir(...)`.

takluyver:
I've renamed it, but with a slight difference to your suggestion: `_path_with_destdir`. I think 'within' is potentially confusing, because that could mean stripping destdir off to get a relative path within that directory.

pradyunsg:
I like your tweak better! :)

src/installer/destinations.py
```diff
@@ -176,20 +208,32 @@ def write_script(
...
+            return
+
+        target_path = self._path_with_destdir(scheme, record.path)
+        for level in self.bytecode_optimization_levels:
+            kwargs: Dict[str, Any] = {}
+            if sys.version_info >= (3, 9):  # pragma: no cover
+                kwargs["stripdir"] = str(self.destdir)
+            compileall.compile_file(target_path, optimize=level, **kwargs)
```
mgorny, line 228: Non-blocking comment: perhaps you want to pass `quiet=1` here? It feels a bit weird that the install itself is entirely quiet, yet compileall is very verbose.

src/installer/destinations.py
```diff
@@ -176,20 +208,32 @@ def write_script(
...
+    def _compile_bytecode(self, scheme: Scheme, record: RecordEntry) -> None:
+        """Compile bytecode for a single .py file."""
+        if scheme not in ("purelib", "platlib"):
+            return
+
+        target_path = self._path_with_destdir(scheme, record.path)
+        for level in self.bytecode_optimization_levels:
+            kwargs: Dict[str, Any] = {}
+            if sys.version_info >= (3, 9):  # pragma: no cover
+                kwargs["stripdir"] = str(self.destdir)
```
mgorny, line 227: Not sure if this causes any practical difference but this results in a path without the leading `/` being embedded, i.e. `usr/lib/python...` rather than the usual `/usr/lib/python...`.

FRidh:
I gave this a try in Nixpkgs, and it seems the path to the interpreter is always prefixed.

What we get
```
/nix/store/9pi9l6c3g8hmmgv656zjfkh7hwwxwd5m-python3.9-bootstrapped-build-installer/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/lib/python3.9/site-packages/flit_core/buildapi.py
```

What we should have
```
/nix/store/9pi9l6c3g8hmmgv656zjfkh7hwwxwd5m-python3.9-bootstrapped-build-installer/lib/python3.9/site-packages/flit_core/buildapi.py
```

Building wheels is done with
```
${python.pythonForBuild.interpreter} -m build --wheel "$pkg" --no-isolation --skip-dependency-check
```

and installing using
```
${python.pythonForBuild.interpreter} -m installer "$whl" --destdir $out --compile-bytecode "0"
```

eli-schwartz:
My understanding is that NixOS specifically does not use DESTDIR because every package already has its own unique PREFIX. No idea how you handle sysconfig paths though (which contain the PREFIX).

FRidh:
> My understanding is that NixOS specifically does not use DESTDIR because every package already has its own unique PREFIX. No idea how you handle sysconfig paths though (which contain the PREFIX).

Correct, every package has a prefix.

With pip we use

```
${python.pythonForBuild.interpreter} -m pip install --no-build-isolation --no-index --prefix=$out  --ignore-installed --no-dependencies --no-cache .
```

src/installer/destinations.py
takluyver: Good point, yes, I think `quiet=1` (printing only on errors) is appropriate.

src/installer/destinations.py
takluyver: I think I've worked out a fix for this (using `ddir` instead of `stripdir`), which also means we can make the same call across all supported Python versions. Can you try the latest commit and see if the paths are correct now?

takluyver:
To be 100% clear, `--destdir` is equivalent to the `--root` option in pip, not `--prefix`. The behaviour you show looks correct for that option (though I understand it's not the behaviour you want).

I floated the idea of a prefix option on #66 (the predecessor to this PR), but @FFY00 and @eli-schwartz argued strongly against it. I get the sense they've spent more time than me thinking about these things, so I deferred to them. But I can't see a way to achieve what you want without a prefix option, short of messing around with sysconfig (which seems like a workaround to me).

One other way to approach this could be through the lower-level 'give us explicit target paths' interface (as in #93), and leave it to you to construct paths like `$out/lib/python3.10/site-packages`.

@pradyunsg any preferences?

FRidh:
> To be 100% clear, --destdir is equivalent to the --root option in pip, not --prefix. The behaviour you show looks correct for that option (though I understand it's not the behaviour you want).

Right.

> I floated the idea of a prefix option on #66 (the predecessor to this PR), but @FFY00 and @eli-schwartz argued strongly against it. I get the sense they've spent more time than me thinking about these things, so I deferred to them. But I can't see a way to achieve what you want without a prefix option, short of messing around with sysconfig (which seems like a workaround to me).

My understanding is that @FFY00 indeed suggests adding a sysconfig scheme:

> Which demonstrates the limitations of overwriting the prefix, and would be require a different tool like the one I described above. But if spack is the one shipping the Python interpreter, I would recommend adding a sysconfig scheme instead sweat_smile

Adding a scheme might be a possibility. We now use the `posix_prefix` scheme as is. If we would make a custom scheme, this scheme would be the `posix_prefix` minus the path to the interpreter that ours has. E.g., we now get when running from a "Nixpkgs buildenv" environment (sort of like a venv but then not limited to just Python):
```
{'stdlib': '/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/lib/python3.9',
 'platstdlib': '/nix/store/hidc6byr4gdxpxckkc4a8s9fbvjl05jy-python3-3.9.9-env/lib/python3.9',
 'purelib': '/nix/store/hidc6byr4gdxpxckkc4a8s9fbvjl05jy-python3-3.9.9-env/lib/python3.9/site-packages',
 'platlib': '/nix/store/hidc6byr4gdxpxckkc4a8s9fbvjl05jy-python3-3.9.9-env/lib/python3.9/site-packages',
 'include': '/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/include/python3.9',
 'platinclude': '/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/include/python3.9',
 'scripts': '/nix/store/hidc6byr4gdxpxckkc4a8s9fbvjl05jy-python3-3.9.9-env/bin',
 'data': '/nix/store/hidc6byr4gdxpxckkc4a8s9fbvjl05jy-python3-3.9.9-env'}
```
Just to compare, a virtualenv, constructed from that env, looks like
```
>>> pprint.pprint(sysconfig.get_paths())
{'data': '/tmp/venv/env',
 'include': '/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/include/python3.9',
 'platinclude': '/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/include/python3.9',
 'platlib': '/tmp/venv/env/lib/python3.9/site-packages',
 'platstdlib': '/tmp/venv/env/lib/python3.9',
 'purelib': '/tmp/venv/env/lib/python3.9/site-packages',
 'scripts': '/tmp/venv/env/bin',
 'stdlib': '/nix/store/dn4fwp0yx6nsa85cr20cwvdmg64xwmcy-python3-3.9.9/lib/python3.9'}
>>>
```

I don't know how the interpreter would behave though if we would make this custom prefix the default prefix. It would run with something looking like
```
{'stdlib': 'lib/python3.9',
 'platstdlib': 'lib/python3.9',
 'purelib': 'python3.9/site-packages',
 'platlib': 'site-packages',
 'include': 'include/python3.9',
 'platinclude': 'include/python3.9',
 'scripts': 'bin',
 'data': ''}
```

jameshilliard:
> I floated the idea of a prefix option on #66 (the predecessor to this PR), but @FFY00 and @eli-schwartz argued strongly against it. I get the sense they've spent more time than me thinking about these things, so I deferred to them.

Since `installer` needs to be able to install itself then it can't depend on any packages that would have `installer` in their own dependency tree, otherwise you would have a dependency cycle.

So IMO `installer` really should provide enough cli functionality to install itself into the desired locations without any external dependencies(other than `flit_core` which would be able to install itself in [#511](https://github.com/pypa/flit/pull/511)).

src/installer/destinations.py
mgorny: It looks to be correct now. Thanks!

mgorny:
I've just gotten a report that the install doesn't make scripts (as in normal scripts, not entry points) executable. I'm not sure if this is something to tackle as part of this PR or separately (#90 might be relevant).

Example wheel: https://files.pythonhosted.org/packages/91/8b/fa0a7295cfb6c81586a2ec0d57ef3e8dc8c425f11726f351f13bcf523cda/pkgcore-0.12.9-py3-none-any.whl

takluyver:
It should install the script files to the correct location, but I don't think it treats them specially at all - rewriting shebangs, setting the execute bit for Unix, or making exe wrappers for Windows. Entry points became the de facto standard for installing executable commands, so the old mechanism to include script files doesn't get much love (see e.g. https://github.com/pypa/pip/issues/2596 ).

[The wheel spec](https://packaging.python.org/en/latest/specifications/binary-distribution-format/) is silent, as far as I can see, on whether it's the responsibility of the code building the wheel to set the executable bit, or whether the code installing it should make anything in the scripts directory executable.

In any case, any special handling for script files is orthogonal to adding a CLI (the existing Python API has the same limitation), so I think it should be addressed separately.

takluyver:
(My mistake, I see it does have special handling for setting the shebang. In that case, I suspect #90 will fix the problem you describe)

pradyunsg:
#90 is the right thing for that. :)

mgorny:
@takluyver, could you update your patch following the changes in main? Not sure if rebasing or merging is the right thing for this repo.

takluyver:
I've merged master in for now - Github makes that easy. @pradyunsg I can rebase it if you prefer.

mgorny:
Thank you. I suppose I'll just switch Gentoo to your fork rather than trying to apply this PR as a patch ;-).

pradyunsg:
I'm gonna squash-merge this one, since it's decently atomic on the whole and I don't find it particularly interesting to have the entire history of the evolution of this PR, recorded on `main`.

pradyunsg:
Thanks @takluyver for taking this over the line, @FFY00 for starting us down this path and everyone else for engaging in this discussion! ^>^

takluyver:
Thanks! Just to link it here, I see @FRidh has opened #98 for further discussion about the idea of a prefix option.

PR #96, opened 2022-01-12
Run tests on CPython on Github actions

takluyver:
I noticed on #94 that most of the CI jobs were not actually running the tests! Only the PyPy job was actually running the tests.

It appears that specifying `-s <session1> -s <session2>` for nox doesn't work - only the last session specified runs. It needs to be specified like `-s <session1> <session2>` instead.


pradyunsg:
Thanks @takluyver! ^>^

PR #99, opened 2022-01-25
Refactor path handling to use pathlib.

jameshilliard:
This should help reduce the amount of special casing needed for windows paths.

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -1,9 +1,9 @@
 """Installer CLI."""
 
 import argparse
-import os.path
 import sys
 import sysconfig
+from pathlib import Path
 from typing import Dict, Optional, Sequence
 
 import installer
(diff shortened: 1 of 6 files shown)


pradyunsg:
Thanks for filing this!

I don't think there's much Windows-related special casing in this codebase and, overall, I think the code style that this PR introduces adds more complexity than it adds value in terms of clarity or ease of use. I'm going to close this, to reflect that I don't think this conversion is worthwhile.

jameshilliard:
> I don't think there's much Windows-related special casing in this codebase and, overall, I think the code style that this PR introduces adds more complexity than it adds value in terms of clarity or ease of use.

I guess the main issue is that using `os.path` makes it rather difficult to reason about correctness since it's otherwise unclear what type of path(unix vs os style) the various codepaths actually expect to get passed. Is there a better way to make the path type demarcations clearer?

Isn't `os.path` generally not recommended to be used in favor of pathlib these days in general?

PR #103, opened 2022-02-03
Implement `--prefix` option

### LINKED ISSUE #98: Support for --prefix
Add support for `--prefix`. Note this is not the same as `--destdir` which was discussed in https://github.com/pradyunsg/installer/issues/58 and has been added.

Discussion on `--prefix` in https://github.com/pradyunsg/installer/pull/94#issuecomment-1013876999 and https://github.com/pradyunsg/installer/pull/66. I don't see anyway around needing `--prefix` in Nixpkgs.

A `sysconfig` scheme is static and doesn't provide the flexibility we need, that is, being able to install every package in its own *prefix*.

Alternatively, I could imagine reading a scheme from say a json or toml file when installing.

Something that could work for us:

```py
var = {"installed_base": "$out", "base": "$out", "platbase": "$out", "installed_platbase": "$out"}
# Note there is no `sysconfig.get_default_scheme()`
sysconfig._expand_vars("posix_prefix", var)
```

mgorny:
Implement a --prefix option that can be used to override platbase
and base when expanding the paths from install scheme.  This can be used
to install paths to a layout using another prefix than the one used
to run installer.

Gentoo use case for this is running installer inside a venv to prepare
installation images for the real system.  The nixOS maintainers have
also indicated interest in this feature.

Fixes #98

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -24,6 +24,13 @@ def _get_main_parser() -> argparse.ArgumentParser:
         help="destination directory (prefix to prepend to each file)",
     )
     parser.add_argument(
+        "--prefix",
+        "-p",
+        metavar="path",
+        type=str,
+        help="override prefix to install packages to",
+    )
+    parser.add_argument(
         "--compile-bytecode",
         action="append",
         metavar="level",
@@ -39,12 +46,18 @@ def _get_main_parser() -> argparse.ArgumentParser:
     return parser
 
 
-def _get_scheme_dict(distribution_name: str) -> Dict[str, str]:
+def _get_scheme_dict(
+    distribution_name: str, prefix: Optional[str] = None
+) -> Dict[str, str]:
     """Calculate the scheme dictionary for the current Python environment."""
-    scheme_dict = sysconfig.get_paths()
+    vars = {}
+    if prefix is None:
+        installed_base = sysconfig.get_config_var("base")
+        assert installed_base
+    else:
+        vars["base"] = vars["platbase"] = installed_base = prefix
 
-    installed_base = sysconfig.get_config_var("base")
-    assert installed_base
+    scheme_dict = sysconfig.get_paths(vars=vars)
 
     # calculate 'headers' path, not currently in sysconfig - see
     # https://bugs.python.org/issue44445. This is based on what distutils does.
@@ -72,7 +85,7 @@ def _main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
 
     with WheelFile.open(args.wheel) as source:
         destination = SchemeDictionaryDestination(
-            scheme_dict=_get_scheme_dict(source.distribution),
+            scheme_dict=_get_scheme_dict(source.distribution, prefix=args.prefix),
             interpreter=sys.executable,
             script_kind=get_launcher_kind(),
             bytecode_optimization_levels=bytecode_levels,
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,3 +1,5 @@
+import os
+
 from installer.__main__ import _get_scheme_dict as get_scheme_dict
 from installer.__main__ import _main as main
 
@@ -7,6 +9,14 @@ def test_get_scheme_dict():
     assert set(d.keys()) >= {"purelib", "platlib", "headers", "scripts", "data"}
 
 
+def test_get_scheme_dict_prefix():
+    d = get_scheme_dict(distribution_name="foo", prefix="/foo")
+    for key in ("purelib", "platlib", "headers", "scripts", "data"):
+        assert d[key].startswith(
+            f"{os.sep}foo"
+        ), f"{key} does not start with /foo: {d[key]}"
+
+
 def test_main(fancy_wheel, tmp_path):
     destdir = tmp_path / "dest"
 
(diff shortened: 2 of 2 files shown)


FRidh (approved):

src/installer/__main__.py
```diff
@@ -39,16 +46,27 @@ def main_parser() -> argparse.ArgumentParser:
...
     # https://bugs.python.org/issue44445. This is based on what distutils does.
     # TODO: figure out original vs normalised distribution names
     scheme_dict["headers"] = os.path.join(
         sysconfig.get_path(
-            "include", vars={"installed_base": sysconfig.get_config_var("base")}
+            "include",
+            vars={
+                "installed_base": prefix
+                if prefix is not None
+                else sysconfig.get_config_var("base")
```
pradyunsg, line 68: Instead of having a conditional inline on an already deeply nested block of code, could you move this out to a local that's assigned in two branches of an if-else statement?

That way the line coverage will actually require covering both possible branches here in the tests.

pradyunsg (commented):
Seems good to me, with one blocking suggestion.

src/installer/__main__.py
mgorny: Done!

mgorny:
Also rebased and reverse the `if` branches to make mypy happier.

adamjstewart:
This solution won't work for macOS framework installations of Python (pretty-print added by me):
```python
$ /usr/bin/python3
>>> import sysconfig
>>> sysconfig.get_paths()
{
    'stdlib': '/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8', 
    'platstdlib': '/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8', 
    'purelib': '/Library/Python/3.8/site-packages', 
    'platlib': '/Library/Python/3.8/site-packages', 
    'include': '/Library/Python/3.8/include', 
    'platinclude': '/Library/Python/3.8/include', 
    'scripts': '/usr/local/bin', 
    'data': '/Library/Python/3.8'
}
>>> sysconfig.get_paths(vars={'base': '/foo/bar', 'platbase': '/foo/bar'})
{
    'stdlib': '/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8', 
    'platstdlib': '/foo/bar/lib/python3.8', 
    'purelib': '/Library/Python/3.8/site-packages', 
    'platlib': '/Library/Python/3.8/site-packages', 
    'include': '/Library/Python/3.8/include', 
    'platinclude': '/Library/Python/3.8/include', 
    'scripts': '/usr/local/bin', 
    'data': '/Library/Python/3.8'
}
```
As you can see, purelib/platlib haven't changed. It looks like they don't use base/platbase in the default scheme:
```python
>>> sysconfig.get_paths(expand=False)
{
    'stdlib': '{installed_base}/lib/python{py_version_short}', 
    'platstdlib': '{platbase}/lib/python{py_version_short}', 
    'purelib': '/Library/Python/{py_version_short}/site-packages', 
    'platlib': '/Library/Python/{py_version_short}/site-packages', 
    'include': '/Library/Python/{py_version_short}{abiflags}/include', 
    'platinclude': '/Library/Python/{py_version_short}{abiflags}/include', 
    'scripts': '/usr/local/bin', 
    'data': '/Library/Python/{py_version_short}'
}
```
When I `pip install --prefix` a package, it doesn't use this scheme at all, it seems to use something like the posix_path scheme (`/foo/bar/lib/pythonX.Y/site-packages`). Does anyone know how pip does this if not by modifying these vars? Maybe they explicitly choose a scheme to use?

adamjstewart:
Dug into [how pip handles this](https://discuss.python.org/t/understanding-site-packages-directories/12959/17?u=adamjstewart). Basically, the solution proposed in this PR is about as good as it will get, and is basically [exactly what pip does](https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/_sysconfig.py#L218). The issue I mentioned above is alluded to in the [pip source code](https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/__init__.py#L486). Pip still uses distutils for Python 3.9 and older, and distutils doesn't have this bug, which is why I didn't encounter it. Obviously we don't want to use distutils in installer, so I think the correct solution is to merge this PR as is and acknowledge that things will be broken for the XCode Python on macOS.

uranusjr:
> distutils doesn't have this bug

Nitpick, a more correct phrasing of this would be _Apple patches distutils correctly, but not sysconfig_ 🙂

mgorny:
FTR, I've ended up doing something a bit different for Gentoo — calling `sysconfig.get_paths()` with `base` and `platbase` overrides: https://github.com/mgorny/gpep517/blob/master/gpep517/__main__.py#L103

mgorny:
> FTR, I've ended up doing something a bit different for Gentoo — calling `sysconfig.get_paths()` with `base` and `platbase` overrides: https://github.com/mgorny/gpep517/blob/master/gpep517/__main__.py#L103

Sorry, I meant different from my original idea of patching `sys.prefix`. Let's disregard my comment, I suppose it just adds noise.

FFY00:
I don't think the way Apple patches sysconfig is necessarily incorrect. It just highlights one of the weaknesses of this proposed option, it does not follow the assumption that all paths will be based on `{prefix}`, this can be because the schemes simply do not use these variables, or that the paths are not based on `{prefix}` at all. Like I said many times before, I don't think adding this is a good idea, but I understand it would be the simplest to fix the issue people are facing, it's just that it will often break unpredictably.

adamjstewart:
FWIW, what I'm planning on doing in Spack is to:

1. Run `get_paths` and `get_config_vars` with the default vars
2. If `purelib`/`platlib`/`include` are prefixed by `base`/`platbase`/`installed_base`/`installed_platbase`, replace those
3. If not, hard-code to `--prefix` + `lib/pythonX.Y/site-package` since that is the most common

I'm hoping that works for most situations. 1 is [how sysconfig is used by pip](https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/_sysconfig.py#L217), and 2 is [how distutils does things](https://github.com/python/cpython/blob/v3.10.4/Lib/distutils/sysconfig.py#L309).

lazka:
Could this be rebased maybe?

mgorny:
> Could this be rebased maybe?

Done.

pradyunsg (approved):

pradyunsg:
I'll probably bring the location logic in sync with what pip's doing with sysconfig in 22.2.

PR #110, opened 2022-03-13
Move away from `import installer.xyz` style imports

pradyunsg:
These are incompatible with import-rewriting mechanisms for vendoring
existing code.


uranusjr:
Why not use relative imports?

pradyunsg:
Well, the codebase largely uses full absolute imports everywhere; so I didn't want to deviate from it in just this one instance.

PR #115, opened 2022-03-29
Fix the unbound is_executable

frostming:
The `is_executable` is from the last for-loop and is unbound in adding `additional_metadata`. Fix it to `False` as additional metadata are likely not executable.

### DIFF
--- a/src/installer/_core.py
+++ b/src/installer/_core.py
@@ -123,7 +123,7 @@ def install(
                 scheme=root_scheme,
                 path=path,
                 stream=other_stream,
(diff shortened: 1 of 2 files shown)


pradyunsg (approved):

pradyunsg:
Would you be willing to add a test case for this, to make sure that record is not executable, even if all the other files are?

frostming:
> Would you be willing to add a test case for this, to make sure that record is not executable, even if all the other files are?

Sure, done.

PR #130, opened 2022-08-01
feat: Switch project metadata to PEP 621

frostming:


pradyunsg:
Was this an outdated main on your end? 🙃

frostming:
Exactly :)

Issue #134, opened 2022-08-15
Can't install wheels with non-normalized names


stefanor:
Names are parsed from the distribution filename, but never normalized.

If a distribution has a non-normalized filename, like `Quart-0.17.0-py3-none-any.whl`, then installation fails:

```Traceback (most recent call last):
  File "/usr/bin/pybuild", line 367, in main
    run(func, i, version, c)
  File "/usr/bin/pybuild", line 317, in run
    result = func(context, args)
  File "/usr/share/dh-python/dhpython/build/plugin_pyproject.py", line 102, in build
    self.build_step2(context, args)
  File "/usr/share/dh-python/dhpython/build/plugin_pyproject.py", line 151, in build_step2
    install(
  File "/usr/lib/python3/dist-packages/installer/_core.py", line 77, in install
    root_scheme = _process_WHEEL_file(source)
  File "/usr/lib/python3/dist-packages/installer/_core.py", line 21, in _process_WHEEL_file
    stream = source.read_dist_info("WHEEL")
  File "/usr/lib/python3/dist-packages/installer/sources.py", line 139, in read_dist_info
    return self._zipfile.read(path).decode("utf-8")
  File "/usr/lib/python3.10/zipfile.py", line 1475, in read
    with self.open(name, "r", pwd) as fp:
  File "/usr/lib/python3.10/zipfile.py", line 1514, in open
    zinfo = self.getinfo(name)
  File "/usr/lib/python3.10/zipfile.py", line 1441, in getinfo
    raise KeyError(
KeyError: "There is no item named 'Quart-0.17.0.dist-info/WHEEL' in the archive"
```

From: https://bugs.debian.org/1008606
Part of the bigger issue in #97.

dimaqq:
I've played a bit with `poetry==1.4.0` and I'm pretty sure it uses `installer==0.6.0`, the latest released.

Yet the above, linked error still happens.

It seems that the fix was not released?

baurt:
This was closed about a week after the `0.6.0` release was cut, so it would make sense that it doesn't work.

> I've played a bit with `poetry==1.4.0` and I'm pretty sure it uses `installer==0.6.0`, the latest released.
> 
> Yet the above, linked error still happens.
> 
> It seems that the fix was not released?

This was closed about a week after the `0.6.0` release was cut, so it would make sense that it doesn't work.

PR #135, opened 2022-08-15
Normalize wheel filenames

### LINKED ISSUE #97: Resolve normalised vs original distribution names
At present, the distribution name comes from the wheel filename, which can be normalised according to the new rule (PEP 503 normalisation, but with `_` in place of `-`) or according to an older rule (any character except `A-Za-z0-9.` is converted to `_`).

If we want the original distribution name for anything, we should get that from wheel metadata. If we want the normalised name, we should normalise it ourselves for consistency.

stefanor:
This is a fairly simple solution to #134, without any refactoring needed to close #97.

There are other places this normalization could be done, but this seems like the logical place to do it. It does mean that you can't get back from a `WheelFilename` object to the filename on disk, but that doesn't seem very important.

Fixes: #134

### DIFF
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -109,7 +109,9 @@ def parse_wheel_filename(filename: str) -> WheelFilename:
     wheel_info = _WHEEL_FILENAME_REGEX.match(filename)
     if not wheel_info:
         raise ValueError(f"Not a valid wheel filename: {filename}")
-    return WheelFilename(*wheel_info.groups())
+    parsed = wheel_info.groups()
+    normalized_name = normalize_distribution_name(parsed[0])
+    return WheelFilename(normalized_name, *parsed[1:])
 
 
 def copyfileobj_with_hashing(
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -93,6 +93,14 @@ def parse_metadata_file(contents: str) -> Message:
     return feed_parser.close()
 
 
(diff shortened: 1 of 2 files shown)


src/installer/utils.py
```diff
@@ -101,7 +109,9 @@ def parse_wheel_filename(filename: str) -> WheelFilename:
     wheel_info = _WHEEL_FILENAME_REGEX.match(filename)
     if not wheel_info:
         raise ValueError(f"Not a valid wheel filename: {filename}")
-    return WheelFilename(*wheel_info.groups())
+    parsed = wheel_info.groups()
+    normalized_name = normalize_distribution_name(parsed[0])
```
pradyunsg, line 113: Sadly, we don't know for sure if the name inside the wheel will be canonicalized.

For example, `zope.interface`'s wheels contain `zope.interface-5.4.0.dist-info` (and not `zope-interface`) thanks, in no small part, to setuptools' behaviour in this area.

pradyunsg: Oh, wait, that's what this is doing. :)

stefanor: So, we're going to either need to probe a few names, or search the directory listing.

stefanor: I see pip searches the zip for a `.dist-info`. So I guess we should do that.

pradyunsg: I agree. Let's close this out, since we've established that merely canonicalising the name isn't sufficient.

PR #137, opened 2022-08-16
Search wheels for `.dist-info` directories

stefanor:
Some wheels don't use normalized names for their `.dist-info` directories, so search the wheel for them.

Fixes: #134

### DIFF
--- a/src/installer/sources.py
+++ b/src/installer/sources.py
@@ -8,7 +8,7 @@ from contextlib import contextmanager
 from typing import BinaryIO, ClassVar, Iterator, List, Tuple, Type, cast
 
 from installer.records import RecordEntry, parse_record_file
-from installer.utils import parse_wheel_filename
+from installer.utils import canonicalize_name, parse_wheel_filename
 
 WheelContentElement = Tuple[Tuple[str, str, str], BinaryIO, bool]
 
@@ -155,10 +155,21 @@ class WheelFile(WheelSource):
             dist_infos = [
                 name for name in top_level_directories if name.endswith(".dist-info")
             ]
+
             assert (
                 len(dist_infos) == 1
             ), "Wheel doesn't contain exactly one .dist-info directory"
-            self._dist_info_dir = dist_infos[0]
+            dist_info_dir = dist_infos[0]
+
+            # NAME-VER.dist-info
+            di_dname = dist_info_dir.rsplit("-", 2)[0]
+            norm_di_dname = canonicalize_name(di_dname)
+            norm_file_dname = canonicalize_name(self.distribution)
+            assert (
+                norm_di_dname == norm_file_dname
+            ), "Wheel .dist-info directory doesn't match wheel filename"
+
+            self._dist_info_dir = dist_info_dir
         return self._dist_info_dir
 
     @property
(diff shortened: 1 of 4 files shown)


src/installer/sources.py
```diff
@@ -122,6 +122,33 @@ def open(cls, path: "os.PathLike[str]") -> Iterator["WheelFile"]:
...
+            ), "Wheel doesn't contain exactly one .dist-info directory"
+            dist_info_dir = dist_infos[0]
+
+            # NAME-VER.dist-info
+            di_dname = dist_info_dir.rsplit("-", 2)[0]
+            norm_di_dname = normalize_distribution_name(di_dname)
+            norm_file_dname = normalize_distribution_name(self.distribution)
+            assert (
+                norm_di_dname == norm_file_dname
+            ), "Wheel .dist-info directory doesn't match wheel filename"
```
stefanor, line 147: This isn't exactly the same check that pip does. It knows what distribution name it's expecting to install, so that's what it's looking for.

In the wheels I've seen that we hit this bug on, this assertion would be true.

ArmaanT:
Is there anything I can do to help get this merged? I've been running into this problem when using bazel to install wheels with capitalization issues.

stefanor:
Ping?

src/installer/sources.py
pradyunsg: We also know the name, via `self.distribution`.

pradyunsg: Ah, right, in case of a mismatch. Nvm me.

src/installer/utils.py
```diff
@@ -93,6 +93,14 @@ def parse_metadata_file(contents: str) -> Message:
     return feed_parser.close()
 
 
+def normalize_distribution_name(name: str) -> str:
```
pradyunsg, line 96: Hmm... this might be easier to parse for folks if this has the same name as the `packaging` function.

```suggestion
def canonicalize_name(name: str) -> str:
```

src/installer/sources.py
```diff
@@ -122,6 +122,33 @@ def open(cls, path: "os.PathLike[str]") -> Iterator["WheelFile"]:
...
+
+            assert (
+                len(dist_infos) == 1
+            ), "Wheel doesn't contain exactly one .dist-info directory"
+            dist_info_dir = dist_infos[0]
+
+            # NAME-VER.dist-info
+            di_dname = dist_info_dir.rsplit("-", 2)[0]
+            norm_di_dname = normalize_distribution_name(di_dname)
+            norm_file_dname = normalize_distribution_name(self.distribution)
```
pradyunsg, line 144:
```suggestion
            norm_di_dname = canonicalize_name(di_dname)
            norm_file_dname = canonicalize_name(self.distribution)
```

pradyunsg (dismissed):
LGTM, with one non-blocking suggestion.

Lemme know if you'd prefer that I merge without that change. :)

pradyunsg:
TestWheelFile.test_finds_dist_info is failing on 3.7.

stefanor:
Applied your name suggestion, and rebased on top of master.

> TestWheelFile.test_finds_dist_info is failing on 3.7.

Added a workaround for that in a separate commit, that's easy to revert later.

pradyunsg (approved):

pradyunsg:
Thanks @stefanor!

PS: I've filed #157 for a few follow up changes that I'd like to make. :)

Issue #120, opened 2022-04-24
[feature request] option to specify installation scheme


greyltc:
From using this in Linux, it seems like the installation scheme used here is always "posix_prefix." I think it would be helpful to have the option to use "posix_user" too!

FFY00:
This also plays into vendor layout customizations, they could add their layouts to Python and then select them with this option.

pradyunsg:
This should already be possible via the API, which takes a bundle of paths for initialising the installation scheme.

Could you be a bit more specific about what you are trying to do, and what you’re requesting here?

uranusjr:
I’m just guessing but maybe the goal is to install packages to the user site with the CLI. Currently `python -m installer --destdir ~/.local` does not do the “right thing”.

FFY00:
Yes, that is not the goal of destdir. I think this would look something like `python -m installer --scheme posix_user`.

uranusjr:
I’m more inclined to think `installer`’s CLI shouldn’t get into alternative schemes at all, and should probably remove `--destdir` altogether. But if y’all do want to do this, it’d be best to rethink how context var override should work (what `--destdir` currently do but only partially).

Right now only `base` can be overridden (via `--destdir`), but if you support changing the scheme, people will eventually want at least `userbase` and ` platbase`, and eventually you’d get into the same situation of pip where people want to use the CLI for all sorts of things like populating an environment on Windows running Python 3.10 to use on a Repsberry Pi with Python 3.6.

eli-schwartz:
Base is never overridden by --destdir as they do conceptually different things.

e.g.
```
'purelib': '{base}/lib/python{py_version_short}/site-packages'
```

Specifically, base for a system installation is /usr, so that fills out as

```
/usr/lib/python3.10/site-packages
```

destdir=./foo prepends to that, so you end up with
```
./foo/usr/lib/python3.10/site-packages
```

And then you pack up the ./foo directory into a system image and install it as an overlay onto / so that the final product gets files in /usr/lib/python3.10/site-packages

> Currently `python -m installer --destdir ~/.local` does not do the “right thing”.

This is a hard command-line API violation, based on the incorrect notion that destdir is intended to override `{base}`. It doesn't do the right thing, because as an API violation it produces garbled output.


[...]

Issue #145, opened 2022-11-09
Expose a utility function to use for sysconfi/distutils paths+scheme handling like oip


pradyunsg:
xref: https://github.com/pypa/pip/issues/11585

Basically, the idea is to mimic the logic in pip's locations subpackage, and expose that via a single function to end users.

FFY00:
Does the locations subpackage do anything other than handling vendor patching?

pradyunsg:
Yes.

See https://github.com/pypa/pip/blob/90f51db1a32592430f2e4f6fbb9efa7a3a249423/src/pip/_internal/locations/_sysconfig.py#L124 and the distutils equivalent for some of the details.

There's also logic to handle the underlying sysconfig/distutils migration, which may be worth mirroring here (depending on how complex that would be).

pradyunsg:
From #146, we should probably also expose this in the CLI.

tho-:
This would just be a matter of adding a CLI `scheme` argument and passing `scheme=scheme` to `sysconfig.get_path` in `__main__.py`. That would have saved me from writing a small wrapper script to do this ...

Any updates on this?

PR #146, opened 2022-11-11
add --scheme argument

greyltc:
solves #120

This adds a new optional argument, `--scheme`, that will force the installation to be done using a specific scheme instead of the default one.

example usage:
```
python -m installer --scheme posix_user some.whl
```

see https://docs.python.org/3/library/sysconfig.html?highlight=installation%20scheme#installation-paths
for a list of valid schemes

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -31,6 +31,13 @@ def _get_main_parser() -> argparse.ArgumentParser:
         help="override prefix to install packages to",
     )
     parser.add_argument(
+        "--scheme",
+        "-s",
+        metavar="scheme",
+        type=str,
+        help="override the default installation scheme",
+    )
+    parser.add_argument(
         "--compile-bytecode",
         action="append",
         metavar="level",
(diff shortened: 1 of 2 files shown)


greyltc:
The failure here seems to be caused by a test coverage check issue.

Why should this block not pass if there are tests for 3.7, 3.8, 3.9 and 3.10?
```python
        if sys.version_info >= (3, 10):
            scheme = sysconfig.get_default_scheme()
        else:
            scheme = sysconfig._get_default_scheme()
```

greyltc:
Is there anything more I can do to help get this merged?

greyltc:
The most obvious thing that's missing here is documentation. argparse's `--help` should do a fine job of explaining the new option, but I haven't touched anything in docs/

pradyunsg:
Hmm... this seems be related to https://github.com/pypa/installer/issues/145 as well; which is the direction I'd prefer to go TBH -- bringing over the existing "battle-tested" code from pip into this library/CLI and exposing it.

pradyunsg:
So... I think going down the route of #145 will either supercede this or be a superset of this change -- and making progress toward that seems like a more sustainable way to evolve overall ("use the same logic as pip" has a strong appeal to it).

To that end, I think I'll close this out -- not because this isn't a complete and correct change on it's own (thanks @greyltc for doing this, being patient and making this really easy to review!) but because I'd like to avoid adding this to the public CLI and restricting this project from potentially being locked out of the ability to pull in behaviours+logic from pip itself.

Issue #105, opened 2022-02-16
Implement foundational validation logic


pradyunsg:
This would be something like:

```
installer.validators.validate_all_files_match_record(source: WheelSource) -> None
installer.validators.validate_record(source: WheelSource) -> None
```

These raise `installer.validators.ValidationError` if the source is not a valid wheel.

pradyunsg:
Closing this since #147 did the basic and #156 covers the rest needed around this.

Issue #100, opened 2022-01-26
Handle wheels whose RECORD file doesn't match


frostming:
I understand it is the packagers' fault that released a wheel with a mismatching RECORD file. But since `pip` isn't that strict for this people will keep complaining "I can't install xxx but pip works fine." At least `installer` could provide an option to skip the RECORD validation.

Some examples:

- `selenium==4.1.0`
- `catboost==1.0.4`

The errors are like:
```
File "C:\Users\user\AppData\Roaming\pdm\venv\lib\site-packages\pdm\installers\installers.py", line 280, in _install_wheel
  File "C:\Users\user\AppData\Roaming\pdm\venv\lib\site-packages\installer\sources.py", line 170, in get_contents
    assert record is not None, "In {}, {} is not mentioned in RECORD".format(
AssertionError: In C:\Users\user\AppData\Local\Temp\pip-unpack-a3gqvydj\catboost-1.0.4-cp39-none-win_amd64.whl, catboost/core.py is not mentioned in RECORD
```

pradyunsg:
Well, of course pip just accepts such wheels work. :)

Longer term, I think the right thing to do here is for pip to become strict on this front. That isn't happening over the weekend tho. I think adding a `yield_files_missing_in_record` flag[^1] that changes this into fabricating a `RecordEntry` with no size or hash is a reasonable idea.

Combined with #105, I think this is a reasonable thing to do.

[^1]: Let's defer bikeshedding on the name please. 😅

PR #147, opened 2022-11-13
Separate validation of `RECORD`

BlueGlassBlock:
This PR adds `validate_record` method to `WheelSource` and a boolean parameter `validate_record` to `install` function.

Currently, `validate_record` implementation of `WheelFile` checks the presence of `RECORD`, entries of files in the wheel, and whether there's a corresponding hash for the entry. (`RECORD.p7s` and `RECORD.jwt` are ignored from [here](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-directory))

`validate_record` parameter of `install` indicates whether `WheelSource.validate_record` will be called during installation.

I've already modified the `fancy_wheel` builder from `conftest.py` to include hashes (and sizes) of files.

The names and places are quite different from #100 and #105, please inform me about what I need to change.

Please review and request changes so that I can push this PR forward 😄 

~~The line ending seems messed up, so please squash when it's ready for merge~~


pradyunsg:
Thanks for this PR @BlueGlassBlock, and appreciate your patience on this.

I'm doing a brain-dump TBH, so apologies for the somewhat unpolished phrasing.

Instintively, I feel like it's valuable to have validation live _outside_ of `WheelSource`, keeping the the actual "details" of how to perform the validation outside of the wheel itself. But, let's check whether that's valid. The motivation for that was to let the tool that holds the wheel to not rely on `installer`. `get_contents` already requires the ability to figure out the relevant `RECORD` for a file entry, so `WheelSource` already needs that. `wheel` even has a function meant for this implemented on their end, so... ok... this makes sense as a method in `WheelSource`. It still shouldn't raise an error defined within from `installer` tho -- at best, this should be `WheelSource.validation_error` (like PEP 517's `UnsupportedOperation`).

pradyunsg (changes requested):
Thanks for filing this PR @BlueGlassBlock! I really appreciate it.

I do have a few suggestions.

I *think* this is my first review on a PR by you so I'll clarify that any comment starting with `nit:` is something that I don't consider being worth an extended discussion. It's more of a nit-pick and I don't think it's worth having an extended discussion on that -- if you disagree with my suggestion or don't really wanna bother with that suggestion, please feel welcome to mark it as resolved without a discussion. :)

tests/test_sources.py
```diff
@@ -92,3 +95,56 @@ def test_provides_correct_contents(self, fancy_wheel):
...
+                    manipulation_func(files, file, archive.read(file))
+                    continue
+
+                files[file] = archive.read(file)
+        # Replace original archive
+        with zipfile.ZipFile(fancy_wheel, mode="w") as archive:
+            for name, content in files.items():
+                archive.writestr(name, content)
+
+    def test_validation_error_no_record(self, fancy_wheel):
```
pradyunsg, line 119: nit: I'd like to follow the current naming style:

```suggestion
    def test_rejects_no_record_on_validate(self, fancy_wheel):
```

```py
    def test_rejects_missing_hash_on_validate(self, fancy_wheel):
```

src/installer/sources.py
```diff
@@ -138,6 +157,44 @@ def read_dist_info(self, filename: str) -> str:
         path = posixpath.join(self.dist_info_dir, filename)
         return self._zipfile.read(path).decode("utf-8")
 
+    def validate_record(self) -> None:
```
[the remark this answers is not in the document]
BlueGlassBlock: `RecordEntry(...).validate(...)` seems a little bit expensive (as it basically requires extracting the whole wheel file instead of just accessing info and `RECORD`) Maybe allowing specifying whether to validate files within `RECORD` is okay?

BlueGlassBlock:
Modified quite a bit. Please review and tell me what to change :)

src/installer/_core.py
```diff
@@ -72,7 +72,6 @@ def install(
     :param destination: where to write the wheel.
     :param additional_metadata: additional metadata files to generate, usually
                                 generated by the caller.
-
```
pradyunsg, line 75: There's an unrelated change here. :)

pradyunsg (commented):
One last thing!

BlueGlassBlock:
Synced and changed. :)

BlueGlassBlock:
Oh, maybe the CLI should be updated too? But I think that should be separated into another PR.

pradyunsg:
Could you remove [14cefe1](https://github.com/pypa/installer/pull/147/commits/14cefe1437b0e2aaf60c677eb7fc0023168a2ff2)? I'd prefer to handle that in a separate PR. :)

BlueGlassBlock:
Force pushed. That commit is saved on another branch.

pradyunsg (approved):

PR #152, opened 2022-11-30
Normalize RECORD file paths when parsing

### LINKED ISSUE #149: PyQt5_Qt5 5.15.2 fails to install: PyQt5_Qt5-5.15.2.dist-info/RECORD is not mentioned in RECORD
The wheel file for `PyQt5_Qt5` version 5.15.2 on PyPi has a RECORD file that lists itself at a a path of `PyQt5_Qt5-5.15.2.dist-info\RECORD`. It is my understanding that 5d1bab41e8c98beffb65e2a3e088a7e8491e2e63 should have fixed this issue, but perhaps there are other places where the path is being handled as per [PEP376](https://peps.python.org/pep-0376/#record) instead of the [Core Metadata specs](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file).

Both `installer` 0.5.1 and the latest main branch were tested.

yodaldevoid:
Fixes #149

### DIFF
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -213,5 +213,8 @@ def parse_record_file(rows: Iterable[str]) -> Iterator[Tuple[str, str, str]]:
             )
             raise InvalidRecordEntry(elements=elements, issues=[message])
 
+        # Convert Windows paths to use / for consistency
+        elements[0] = elements[0].replace("\\", "/")
+
         value = cast(Tuple[str, str, str], tuple(elements))
         yield value
--- a/tests/test_records.py
+++ b/tests/test_records.py
@@ -273,3 +273,12 @@ class TestParseRecordFile:
             ),
             ("distribution-1.0.dist-info/RECORD", "", ""),
         ]
+
+    def test_parse_record_entry_with_backslash_path(self):
+        record_lines = [
+            "distribution-1.0.dist-info\\RECORD,,",
+        ]
+        records = list(parse_record_file(record_lines))
+        assert records == [
+            ("distribution-1.0.dist-info/RECORD", "", ""),
+        ]


pradyunsg (changes requested):
Thanks for the PR! Could you add a dedicated test for normalisation, instead of tweaking an existing test?

pradyunsg (approved):

yodaldevoid:
I forgot that `\` is only a valid separator on Windows (as per the [Core Metadata specs](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file)) so I have the test expecting different results based on the environment.

This is fine as the tests are run on both Windows and Linux, but an alternative would be to change `os.sep` in the test in order to test both code paths no matter the system running the tests. Thoughts?

pradyunsg (changes requested):
Hmmm... thinking about this a bit more, I have one suggestion for a change.

src/installer/records.py
```diff
@@ -213,5 +213,9 @@ def parse_record_file(rows: Iterable[str]) -> Iterator[Tuple[str, str, str]]:
             )
             raise InvalidRecordEntry(elements=elements, issues=[message])
 
+        # Convert Windows paths to use / for consistency
+        if os.sep == "\\":
+            elements[0] = elements[0].replace("\\", "/")  # pragma: no cover
```
pradyunsg, line 218:
```suggestion
        # Convert Windows paths to use / for consistency
        elements[0] = elements[0].replace("\\", "/")
```

It'd be better to consistently replace `\` with `/`, regardless of OS. Otherwise, a wheel built on Windows won't be installable on other platforms; which is what we want to solve here.

pradyunsg (approved):

pradyunsg:
I've amended the commit message on this. Thanks @yodaldevoid for this PR! ^>^

Issue #156, opened 2022-12-13
Make follow-up improvements to `validate_record`


pradyunsg:
There's also a few more changes to make around this (as a follow up to #105):

- rename the argument to `validate_contents`, and make it kw-only.
- should validate that everything has a hash/size even if `validate_contents` is False.
- should validate that RECORD signatures are _not_ mentioned in the RECORD.
- should validate that the RECORD file is mentioned without hash or size.
- documentation needs to present an example and add cross-references to/from `WheelSource.validation_error`.

pradyunsg:
FWIW, I'm also not a big fan of the inconsistency within `test_sources.py` around function tests vs method tests and would prefer to unify things under the `TestWheelFile` class.

pradyunsg:
Thanks @BlueGlassBlock! You're awesome! ❤️

PR #161, opened 2022-12-27
Add `--validate-record` to CLI

BlueGlassBlock:
Corresponding to #147.

`python -m installer --validate-record=all pkg.whl`

Choices are `all` `entries` (only name precedence in RECORD) and `none` (default)

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -43,6 +43,14 @@ def _get_main_parser() -> argparse.ArgumentParser:
         action="store_true",
         help="don't generate bytecode for installed modules",
     )
+    parser.add_argument(
+        "--validate-record",
+        metavar="part",
+        default="none",
+        type=str,
+        choices=["all", "entries", "none"],
+        help="validate the wheel against certain part of its record (default=none)",
+    )
     return parser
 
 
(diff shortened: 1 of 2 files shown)


pradyunsg:
Adding a cross-reference to https://github.com/pypa/installer/issues/156 :)

pradyunsg:
Thanks @BlueGlassBlock! ^>^

src/installer/__main__.py
```diff
@@ -84,6 +92,8 @@ def _main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
         bytecode_levels = [0, 1]
 
     with WheelFile.open(args.wheel) as source:
+        if args.validate_record != "none":
+            source.validate_record(args.validate_record == "all")
```
dlqqq, line 96: Did you forget to handle the `"entries"` case here? Right now, it seems the current behavior of `--validate-record=entries` is the same as `--validate-record=none`.

BlueGlassBlock: When passing `entries` instead of `none`, basic record format validations will be performed (as of #156), instead computing and comparing file hashes (which may cause perf issues when the files are large).

BlueGlassBlock: `WheelFile.validate_record()`'s default is validating hashes though.

PR #163, opened 2023-01-12
Explicitly use policy=compat32

hauntsaninja:
The documentation in
https://packaging.python.org/en/latest/specifications/core-metadata/ claims that metadata should be parsed with compat32 policy.

The documentation in
https://docs.python.org/3/library/email.parser.html#feedparser-api claims that policy should always be specified and that the default will change at some point. When CPython changes the default, that would presumably introduce a bug here.

### DIFF
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -12,6 +12,7 @@ from collections import namedtuple
 from configparser import ConfigParser
 from email.message import Message
 from email.parser import FeedParser
+from email.policy import compat32
 from typing import (
     TYPE_CHECKING,
     BinaryIO,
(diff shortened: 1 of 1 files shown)


abravalheri:
@hauntsaninja, do you know if `compat32` is compatible with UTF-8?

hauntsaninja:
I'm not sure, although given that FeedParser takes a str, not bytes, I think it might have to do something weird to not handle UTF-8 (and from a glance at the CPython code I'm not seeing anything weird).

Note that this PR doesn't change the policy, it just makes it explicit (as recommended by the Python documentation, due to future change).

hauntsaninja:
@takluyver added the mention of compat32 to the core metadata spec docs I linked in https://github.com/pypa/packaging.python.org/commit/3c855a61bc095fab8d75daabc5a00be38f60b2df

pradyunsg:
Thanks @hauntsaninja! ^>^

PR #167, opened 2023-03-02
Refactor `validate_record`

BlueGlassBlock:
Address #156

- [x] Moved tests to `TestWheelFile` 
- [x] rename the argument to `validate_contents`, and make it kw-only.
- [x] should validate that everything has a hash/size even if `validate_contents` is False.
- [x] should validate that RECORD signatures are _not_ mentioned in the RECORD.
- [x]  should validate that the RECORD file is mentioned without hash or size.
- [ ]  documentation needs to present an example and add cross-references to/from `WheelSource.validation_error`.

### DIFF
--- a/src/installer/sources.py
+++ b/src/installer/sources.py
@@ -188,27 +188,29 @@ class WheelFile(WheelSource):
         path = posixpath.join(self.dist_info_dir, filename)
         return self._zipfile.read(path).decode("utf-8")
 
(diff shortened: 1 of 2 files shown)


pradyunsg:
Thanks for filing this @BlueGlassBlock! I'll merge this since this gets us most of the way there and we can tackle the docs in a follow up! :)

PR #174, opened 2023-03-19
Pass kwarg for `validate_record` when called from CLI

onerandomusername:
a kwarg was left off of the call in #161


BlueGlassBlock:
Thanks, apparently I forgot to update that PR after #167 was merged >_<

pradyunsg (approved):

PR #179, opened 2023-04-08
Cope with invalid hash algorithms in RECORD

### LINKED ISSUE #178: record validation fails on invalid hashes
as encountered in the wild with swat 1.12.1 and reported at https://github.com/python-poetry/poetry/issues/7778 - `RECORDS` contains lines like

```
swat-1.12.1.dist-info/METADATA,sha=c2717134222c10492eaaea1040f6d2b1cd077054f1e7e4ea4f8b1cc48498282f,7737
```
which (among other things wrong with it) says `sha` where it means `sha256`.

Current behaviour is an uncaught `ValueError` from `hashlib.new()`

Here's a testcase:
```python
    def test_rejects_record_containing_unknown_hash(self, fancy_wheel):
        with WheelFile.open(fancy_wheel) as source:
            record_file_contents = source.read_dist_info("RECORD")

        new_record_file_contents = record_file_contents.replace("sha256=", "sha=")
        replace_file_in_zip(
            fancy_wheel,
            filename="fancy-1.0.0.dist-info/RECORD",
            content=new_record_file_contents,
        )

        with WheelFile.open(fancy_wheel) as source:
            with pytest.raises(
                WheelFile.validation_error,
                match="RECORD file contains invalid hash `sha`",
            ):
                source.validate_record(validate_contents=True)
```

but it's not clear to me where would be the cleanest place to detect or catch this error so I have no fix to offer

dimbleby:
Fixes #178

I said there that I didn't have a fix to offer; but perhaps even a wrong fix is a better starting point - or anyway has more momentum - than no fix at all


dimbleby:
I can't tell what the test failure at https://github.com/pypa/installer/actions/runs/4646834563/jobs/8680555289?pr=179 was complaining about

I pushed again in the hope of kicking a retry but apparently that needs manual approval

dimbleby:
oh this

> FAIL Required test coverage of 100% not reached. Total coverage: 99.82%

I find it hard to believe that windows 3.9 uniquely fails to get to [the line in question](https://github.com/pypa/installer/blob/87f0df5defb8ba6e738096976281460bbd44abe0/src/installer/sources.py#L282), so not sure what to do about that

dimbleby:
I pushed a commit that adds `# pragma: no cover` to that line, and notes that it is a false positive reported only on windows and python < 3.10.  Say if you have a better idea!

src/installer/records.py
```diff
@@ -175,13 +175,15 @@ def from_elements(cls, path: str, hash_: str, size: str) -> "RecordEntry":
...
             issues.append("`path` cannot be empty")
 
+        hash_value: Optional[Hash] = None
         if hash_:
             try:
-                hash_value: Optional[Hash] = Hash.parse(hash_)
+                hash_value = Hash.parse(hash_)
+                if hash_value.name not in hashlib.algorithms_available:
+                    issues.append(f"invalid hash algorithm '{hash_value.name}'")
+                    hash_value = None
```
pradyunsg, line 184: Should this be performing this validation in a validate command, since this would prevent a `RecordEntry` from being created with a hash that isn't in `algorithms_available`?

pradyunsg:
Thanks for filing this PR @dimbleby, and appreicate your patience on this as I went away from OSS for a month.

> Say if you have a better idea!

The only reason I can think of is that `sha` is a valid algorithm name on those (the only check being undertaken is `in algorithms_available` so it has to be in that list/set). Consider renaming the algorithm to `notavailable=` or `invalid=` instead?

dimbleby:
No worries, take your time!

if `sha` was somehow a valid algorithm name then the new test wouldn't pass at all, right?

src/installer/records.py
dimbleby: My thinking was that it's desirable to be unable to create a `RecordEntry` with a bad hash algorithm, but I'm happy to be guided in another direction if you have a preference.

Yet another approach, doubling down on "invalid hash algorithms are invalid" would be to raise a `ValueError` from `Hash.parse()`, though the error message from that path isn't currently very informative

pradyunsg:
It would, if it's conditionally available on only certain platforms. 😬

src/installer/records.py
```diff
@@ -175,13 +175,15 @@ def from_elements(cls, path: str, hash_: str, size: str) -> "RecordEntry":
         if not path:
             issues.append("`path` cannot be empty")
 
+        hash_value: Optional[Hash] = None
         if hash_:
             try:
-                hash_value: Optional[Hash] = Hash.parse(hash_)
+                hash_value = Hash.parse(hash_)
+                if hash_value.name not in hashlib.algorithms_available:
```
eli-schwartz, line 182: PEP 376 / https://packaging.python.org/en/latest/specifications/recording-installed-packages/ requires that RECORD must be in hashlib.algorithms_guaranteed, using _available is not good enough.

But the wheel format itself (https://packaging.python.org/en/latest/specifications/binary-distribution-format/) doesn't say anything one way or another about it...

> RECORD is a list of (almost) all the files in the wheel and their secure hashes. Unlike PEP 376, every file except RECORD, which cannot contain a hash of itself, must include its hash. The hash algorithm must be sha256 or better; specifically, md5 and sha1 are not permitted, as signed wheel files rely on the strong hashes in RECORD to validate the integrity of the archive.

So it disagrees with PEP 376 in at least two points:
- hash all files, even pyc
- md5 and sha1 are algorithms_guaranteed, but still not allowed

And therefore the PEP cannot be taken as authoritative on algorithms_guaranteed. But only for wheels. Once e extracted it does need to be in _guaranteed.

Fun.

...

In theory this would solve the platform-conditional concern.

eli-schwartz: It's not obvious to me what the intended availability of `installer.records` is. Can I use it to parse the RECORD in a wheel file, or one in site-packages? Is the appropriate guard here to enforce this when creating new RECORD files only?

Experimental patch to validate this as part of destinations, where we know that any algorithm passed is definitely intended for creating a new installed RECORD file: 67da026b4953b9e5f5acdb3118cb76913d1a60dc.

dimbleby:
@pradyunsg if `sha` was conditionally available only on certain platforms, then the new test would fail on those certain platforms - yes?  

The test verifies that we get an error complaining that the algorithm name is invalid: I don't see how that test can pass without hitting the line of code in question.  I really don't see how this can be anything other than coverage weirdness.

src/installer/records.py
dimbleby: I'm not sure what conclusions, if any, you reach from the above.  

I am therefore politely saying "noted"... but not doing anything about it!

Please say if you have opinions about how this MR ought to look (or submit an MR, I'm fine with this one being overtaken by something better)

PR #183, opened 2023-05-22
fix: validate wheel files in a RAM friendly way

ralbertazzi:
See https://github.com/python-poetry/poetry/issues/7983

Content validation of a wheel record currently loads the entire file in memory with a `self._zipfile.read(item)`. This is extremely inefficient from big wheels (the well known PyTorch has now >2 GB wheel files) and leads to an extremely high RAM consumption. This PR fixes this behaviour by reading the zip file content in a buffered way, as other parts of the codebase are already doing. Unfortunately this required a small change to some signatures.

### DIFF
--- a/src/installer/sources.py
+++ b/src/installer/sources.py
@@ -287,11 +287,11 @@ class WheelFile(WheelSource):
                     f"In {self._zipfile.filename}, hash / size of {item.filename} is not included in RECORD"
                 )
             if validate_contents:
-                data = self._zipfile.read(item)
-                if not record.validate(data):
-                    issues.append(
-                        f"In {self._zipfile.filename}, hash / size of {item.filename} didn't match RECORD"
-                    )
+                with self._zipfile.open(item, "r") as stream:
+                    if not record.validate(cast("BinaryIO", stream), item.file_size):
+                        issues.append(
+                            f"In {self._zipfile.filename}, hash / size of {item.filename} didn't match RECORD"
+                        )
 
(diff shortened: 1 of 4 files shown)


src/installer/sources.py
```diff
@@ -287,11 +287,11 @@ def validate_record(self, *, validate_contents: bool = True) -> None:
                     f"In {self._zipfile.filename}, hash / size of {item.filename} is not included in RECORD"
                 )
             if validate_contents:
-                data = self._zipfile.read(item)
-                if not record.validate(data):
-                    issues.append(
-                        f"In {self._zipfile.filename}, hash / size of {item.filename} didn't match RECORD"
-                    )
+                with self._zipfile.open(item, "r") as stream:
+                    if not record.validate(cast("BinaryIO", stream), item.file_size):
```
dimbleby, line 291: (straying into changes beyond the scope of this MR: but so far as I can see this repository could use `IO[bytes]` everywhere it currently uses `BinaryIO` which would save some casting)

pradyunsg:
I've filed #185 for this to have an issue associated with the PR; in case there's any high-level details to discuss. There likely aren't but it can't hurt to have an issue to close and drive not-PR-specific discussions into.

Other than that, I don't think making a backwards incompatible change is necessary here -- I've filed #186 that does not contain backwards incompatible API changes and instead adds a new method and deprecates the (problematic) `RecordEntry.validate(data)` method.

PR #186, opened 2023-05-29
Validate RECORD file using streams instead of reading in-memory

### LINKED ISSUE #185: RECORD file validation loads the entire file in memory
x-ref https://github.com/python-poetry/poetry/issues/7983

This results in OOM issues when installing packages with really large files.

pradyunsg:
Supercedes and closes #183 
Fixes #185 

This ensures that the validation can be performed without loading the entire file in-memory.

### DIFF
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -4,7 +4,9 @@ import base64
 import csv
 import hashlib
 import os
-from typing import Iterable, Iterator, Optional, Tuple, cast
+from typing import BinaryIO, Iterable, Iterator, Optional, Tuple, cast
+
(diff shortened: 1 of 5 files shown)


ralbertazzi:
Thank you @pradyunsg , your solution is cleaner :)

Another optimization that we can bring is performing hash validation just once, as we currently perform it during 1. wheel validation 2. wheel installation. I tried hacking it away in https://github.com/python-poetry/poetry/pull/8027 but it would be clearly better to see this natively supported by `installer`. I first want to get a better understanding at the performance benefit and, if any, I will open an issue.

PR #202, opened 2023-11-23
Rename `htmldir` to `dirhtml`  in ReadTheDocs configuration

edgarrmondragon:


.readthedocs.yml
```diff
@@ -1,11 +1,15 @@
 version: 2
 
 sphinx:
-  builder: htmldir
+  builder: dirhtml
```
edgarrmondragon, line 4: <blockquote>
<p>The <code class="docutils literal notranslate"><span class="pre">htmldir</span></code> builder option was renamed to <code class="docutils literal notranslate"><span class="pre">dirhtml</span></code> to use the same name as sphinx.
Configurations using the old name will continue working.</p>
</blockquote>

https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-builder

pradyunsg (approved):

edgarrmondragon:
I've rebased and renamed this PR now that https://github.com/pypa/installer/pull/204 was merged

PR #205, opened 2023-12-02
Use Black mirror for faster pre-commit

hugovk:
The mirror uses a mypyc-compiled wheel which is faster.

https://github.com/psf/black-pre-commit-mirror

Also update some pins whilst we're here.


pradyunsg (approved):
Thank you!

Secrus (approved):

PR #211, opened 2024-01-05
Apply flake8-bugbear rule

DimitriPapadopoulos:
**B904** Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling

### DIFF
--- a/src/installer/scripts.py
+++ b/src/installer/scripts.py
@@ -126,9 +126,9 @@ class Script:
         key = (self.section, kind)
         try:
             name = _ALLOWED_LAUNCHERS[key]
-        except KeyError:
+        except KeyError as exc:
             error = f"{key!r} not in {sorted(_ALLOWED_LAUNCHERS)!r}"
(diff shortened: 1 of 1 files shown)


src/installer/scripts.py
```diff
@@ -126,9 +126,9 @@ def _get_launcher_data(self, kind: "LauncherKind") -> Optional[bytes]:
         key = (self.section, kind)
         try:
             name = _ALLOWED_LAUNCHERS[key]
-        except KeyError:
+        except KeyError as exc:
             error = f"{key!r} not in {sorted(_ALLOWED_LAUNCHERS)!r}"
-            raise InvalidScript(error)
+            raise InvalidScript(error) from exc
```
pradyunsg, line 131:
```suggestion
            raise InvalidScript(error) from None
```

We shouldn't expose the `KeyError` to callers.

pradyunsg (approved):

PR #216, opened 2024-02-13
Add --overwrite-existing option that overwrites existing files

### LINKED ISSUE #215: Add `--force` option
Feature request for a new option, `--force`, that will disregard and overwrite already existing files when installing a package.

I'd be happy to submit a PR for this, unless there is something I haven't considered and this is a bad idea.

## Context

I maintain a number of Python projects in the Arch User Repository (AUR) and for projects that require tests to run on an installed package I use the following pattern:

```sh
build() {
  cd "$_archive"

  python -m build --wheel --no-isolation
}

check() {
  cd "$_archive"

  rm -rf tmp_install
  python -m installer --destdir=tmp_install dist/*.whl

  local site_packages=$(python -c "import site; print(site.getsitepackages()[0])")
  export PYTHONPATH="$PWD/tmp_install/$site_packages"
  pytest
}
```

I.e. to install the package to a temporary location and add that to `PYTHONPATH` before running the tests.

The `rm` invocation is necessary when the directory is not clean and the package has already been installed in the `tmp_install` directory. However, this invocation could be made redundant by `installer` having the option to `--force` the install and disregard/overwrite any existing package files.

carlsmedstad:
Took a stab at implementing what I suggested in #215.

Implement the `--force` option that, if supplied, will make installer overwrite any already existing package files instead of failing. With this flag, installer can be used in an idempotent manner, i.e. the same command can be executed multiple times with the same result:

```sh
python -m installer --force --destdir=tmp dist/*.whl
python -m installer --force --destdir=tmp dist/*.whl
python -m installer --force --destdir=tmp dist/*.whl
```

One other candidate for the option name I have in mind is  `--overwrite` which I think is more mechanically descriptive, but it might be less ubiquitous, as tools like `mv` and `cp` use `-f/--force`.

Resolves #215

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -51,6 +51,11 @@ def _get_main_parser() -> argparse.ArgumentParser:
         choices=["all", "entries", "none"],
         help="validate the wheel against certain part of its record (default=none)",
     )
+    parser.add_argument(
+        "--force",
+        action="store_true",
+        help="silently overwrite existing files",
+    )
     return parser
 
 
@@ -101,6 +106,7 @@ def _main(cli_args: Sequence[str], program: Optional[str] = None) -> None:
                 script_kind=get_launcher_kind(),
                 bytecode_optimization_levels=bytecode_levels,
                 destdir=args.destdir,
+                force=args.force,
             )
             installer.install(source, destination, {})
 
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -52,7 +52,7 @@ def _get_main_parser() -> argparse.ArgumentParser:
         help="validate the wheel against certain part of its record (default=none)",
     )
     parser.add_argument(
-        "--force",
+        "--overwrite-existing",
         action="store_true",
         help="silently overwrite existing files",
     )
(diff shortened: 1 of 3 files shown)


jvolkman:
(I originally made this comment on the attached issue, but meant to make it here)

`--force` is a pretty generic name. There are other causes of installation failures - such as problems with the `RECORD` file - that won't be ignored by `--force`. Maybe rename to something more specific like `--overwrite-existing`?

carlsmedstad:
That's a much better name, thanks. Changed it in 829b4839aaacefa53ae64411b953bba34f22276c.

dougthor42:
Could this be added as an arg to [`install`](https://github.com/pypa/installer/blob/e50e78eaf8c5211aafe0eb4f9ef41d28d8286bf9/src/installer/_core.py#L64) as well?

```python
installer.install(
    source=wheel_source,
    destination=destination,
    additional_metadata={"INSTALLER": "me"},
    overwrite_existing=True,  # New
)
```

dougthor42:
Is there anything that I can do to help get this merged? I'm looking forward to being able to use this in https://github.com/bazelbuild/rules_python

carlsmedstad:
Echoing @dougthor42, is there anything I can do to have this merged and eventually released? Thanks!

Secrus:
@carlsmedstad please rebase your changes and ping me once it's ready and I will review

carlsmedstad:
@Secrus Thank you! Rebase done.

carlsmedstad:
Let me know if you'd like me to squash the commits.

Secrus (approved):

PR #225, opened 2024-06-16
Avoid `coding: utf-8` in Python 3 scripts

DimitriPapadopoulos:
As far as I know, only Python 3 is supported. Therefore, get rid of `coding: utf-8` in generated scripts, as UTF-8 is the default.

Also, it could be that `module` or `import_name` are not ASCII, but then I would rather add tests with non-ASCII module and function names.

### DIFF
--- a/src/installer/scripts.py
+++ b/src/installer/scripts.py
@@ -41,7 +41,6 @@ _ALLOWED_LAUNCHERS: Mapping[Tuple["ScriptSection", "LauncherKind"], str] = {
 }
 
 _SCRIPT_TEMPLATE = """\
-# -*- coding: utf-8 -*-
 import re
 import sys
 from {module} import {import_name}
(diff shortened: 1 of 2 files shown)


Secrus:
This is not really valid, since you can use installer as library and make it install a wheel into Python 2 environment (in which case, IIRC, the coding info is needed).

Secrus:
This is not really valid, since you can use installer as library and make it install a wheel into Python 2 environment (in which case, IIRC, the coding info is needed).

DimitriPapadopoulos:
If Python 2 target environments are supported, then yes, it's best to keep this as is for non-ASCII module and function names.

I may be mistaken, but I cannot find details about support for Python 2 environments in the documentation.

Secrus:
There might not be a mention of that in the docs. No matter that, I would leave the coding comment in place. It doesn't hurt us to keep it there.

PR #230, opened 2024-08-17
Introduce more Ruff rules

Secrus:


src/installer/exceptions.py
```diff
@@ -5,5 +5,5 @@ class InstallerError(Exception):
     """All exceptions raised from this package's code."""
 
 
-class InvalidWheelSource(InstallerError):
+class InvalidWheelSource(InstallerError):  # noqa: N818
```
Secrus, line 8: Since we are doing a major release, I was thinking that we might do a slight API breakage and rename this to `InvalidWheelSourceError`

src/installer/records.py
```diff
@@ -18,7 +19,7 @@
 
 
 @dataclass
-class InvalidRecordEntry(Exception):
+class InvalidRecordEntry(Exception):  # noqa: N818
```
Secrus, line 22: ditto with the API breakage

src/installer/scripts.py
```diff
@@ -87,7 +87,7 @@ def _build_shebang(executable: str, forlauncher: bool) -> bytes:
     return b"#!/bin/sh\n'''exec' " + quoted + b' "$0" "$@"\n' + b"' '''"
 
 
-class InvalidScript(ValueError):
+class InvalidScript(ValueError):  # noqa: N818
```
Secrus, line 90: ditto with the API breakage

tests/test_scripts.py
```diff
@@ -32,12 +32,8 @@ def test_script_generate_space_in_executable():
 def _read_launcher_data(section, kind):
     prefix = {"console": "t", "gui": "w"}[section]
     suffix = {"win-ia32": "32", "win-amd64": "64", "win-arm": "_arm"}[kind]
-    filename = os.path.join(
-        os.path.dirname(os.path.abspath(_scripts.__file__)),
-        f"{prefix}{suffix}.exe",
-    )
-    with open(filename, "rb") as f:
-        return f.read()
+    file = Path(_scripts.__file__).resolve().parent / f"{prefix}{suffix}.exe"
```
pradyunsg, line 35: This now hits the filesystem due to the `.resolve` which... I don't think that's strictly necessary?

pradyunsg (approved):
LGTM, barring one non-blocking concern.

src/installer/exceptions.py
pradyunsg: I don't think this lint rule is worth an API breakage.

If we do rename this, I think it's relatively cheap to add a `InvalidWheelSource = InvalidWheelSourceError`.

Honestly though, I think that name is worse than the current name and there's precedence that exceptions don't need an "error" in their name, like https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired and more in the standard library. "Invalid..." sufficently communicates the situation IMO.

pradyunsg: None the less, if we want to do this, let's do it in a separate PR.

Secrus: Makes sense. Decided to just ignore the rule. I guess we are smart enough to rule our exception names.

tests/test_scripts.py
Secrus: True, didn't notice it before. Removed `.resolve()`, things still work fine.

PR #237, opened 2024-09-10
Apply repo-review rules

DimitriPapadopoulos:
https://learn.scientific-python.org/development/guides/repo-review/?repo=pypa%2Finstaller&branch=main

Rules [`RF101`](https://learn.scientific-python.org/development/guides/style#RF101) and [`RF103`](https://learn.scientific-python.org/development/guides/style#RF103)will be enforced by #230.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ dynamic = [
 
 [tool.ruff]
 fix = true
-src = ["src"]
 extend-exclude = [
     "noxfile.py",
     "docs/*"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -44,7 +43,25 @@ select = [
...
+    "D206",
+    "D300",
+    "Q000",
+    "Q001",
+    "Q002",
+    "Q003",
+    "COM812",
+    "COM819",
+    "ISC001",
+    "ISC002",
```
Secrus, line 63: Do we need those ignores? We don't use those rules; they are not included as part of #230 and are not enabled by default. I don't see a reason for ignoring the rules that are not enabled.

Secrus (approved):

PR #242, opened 2024-10-12
Drop support for Python 3.8

Secrus:
Drop support for Python 3.8 and add a classifier for Python 3.13

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,15 +10,15 @@ readme = "README.md"
 authors = [
     { name = "Pradyun Gedam", email = "pradyunsg@gmail.com" },
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 classifiers = [
     "License :: OSI Approved :: MIT License",
     "Programming Language :: Python :: 3 :: Only",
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -28,6 +28,7 @@ dynamic = [
 "GitHub" = "https://github.com/pypa/installer"
 
 [tool.ruff]
+target-version = "py39"
```
pradyunsg, line 31: Ruff is smart-enough to infer this from requires-python, so let's not set this.

pyproject.toml
```diff
@@ -42,6 +43,7 @@ select = [
     "I",
     "ISC",
     "D",
+    "UP"
```
pradyunsg, line 46: I'd appreciate if we could introduce this rule in a separate commit from the ones bumping the supported Python version window (possibly as a separate commit that's before the rest of the logic).

Not opposed to adding it, just making the changes slightly more atomic.

Secrus: This is added with #230, will remove from this PR.

pradyunsg (approved):

PR #245, opened 2024-10-15
src/installer/utils.py: sort entries before writing out RECORD file

kanavin:
This helps build reproducibility by ensuring the RECORD file is the same regardless of the order of the input files.

### DIFF
--- a/src/installer/utils.py
+++ b/src/installer/utils.py
@@ -222,7 +222,7 @@ def construct_record_file(
         io.BytesIO(), encoding="utf-8", write_through=True, newline=""
     )
     writer = csv.writer(stream, delimiter=",", quotechar='"', lineterminator="\n")
-    for scheme, record in records:
+    for scheme, record in sorted(records, key=lambda x: x[1].path):
         writer.writerow(record.to_row(prefix_for_scheme(scheme)))
     stream.seek(0)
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Note that it is wiser to delete the RECORD file anyway since you do not want to indicate to `pip` that it knows how to uninstall projects it didn't install and which are tracked by a more comprehensive installation database.

(Plans for pip to respect an INSTALLER identity fell through, unfortunately.)

rossburton:
@eli-schwartz I like that argument.  Can you point at prior art in eg gentoo as a rationale for doing it distro-wide?

rossburton:
(and if that's a generally accepted point of view, maybe installer should have an option to not write them for the case where we don't expect anything to remove the packages?)

rossburton:
Either way though, reproducible files are good.

eli-schwartz:
@rossburton https://bugs.gentoo.org/927818

eli-schwartz:
> Note that it is wiser to delete the RECORD file anyway

This is a typo, it should be "it is wise to delete [...] anyway" -- the word "wiser" would imply that this PR should be avoided and replaced by deletion but that wasn't what I was trying to write. :)

Secrus (approved):

Issue #257, opened 2025-01-15
--executable option to override sys.executable


riastradh:
`python -m installer` supports several options for controlling the paths in the output when they don't match the Python system that is running:

- `--destdir` puts everything relative to a system root
- `--prefix` overrides `sys.prefix` for the installation prefix within that system root where files are installed

But there appears to be no way to control the path that is stored in `#!` lines of entry points that `installer` creates.  In contrast, for example, historically with setuptools, `python setup.py install` would respect `[build] executable = '/path/to/python'` in setup.cfg; this functionality doesn't appear to be available through installer.  It would be nice if `python -m installer` additionally supported:

- `--executable` overrides `sys.executable` for the Python executable path within that system root used in `#!` lines generated for entry points

This is important for cross-compilation into a fixed file system layout in the destdir that is independent of where Python happened to be installed on a development machine.  For example, if I set up a virtualenv in /home/riastradh/venv for a cross-compilation toolchain, `sys.executable` will be /home/riastradh/venv/bin/python, but the scripts installed in the destdir need to start with `#!/opt/pkg/bin/python` -- just like files installed in the destdir need to go under /opt/pkg/lib/python3.12/site-packages according to `--prefix`, rather than under /home/riastradh/venv/lib/python3.12/site-packages according to `sys.prefix`.

Issue #107, opened 2022-03-11
Option to set interpreter to use


rossburton:
I'm looking at using installer to install wheels in our distribution.  At runtime `sys.executable` is a special python wrapper that we use, but we don't want to use that on the target as it doesn't exist.

I think adding a simple `--interpreter` option should be sufficient for this use-case to set the `interpreter` parameter on `SchemeDictionaryDestination`, leaving the default as `sys.executable`.

pradyunsg:
`installer`'s CLI is designed to install the wheel, for the currently running Python.

I'm curious, what exactly is your usecase for you're modifying with `sys.executable` and having that not point to a regular Python executable? Also, what project is this?

rossburton:
Yocto Project, we're cross-compiling.  We have a magic sysconfig which knows about target paths when asked and assuming that sys.executable is the target is a bad assumption to make.

The interpreter we want to use is a standard interpreter, it's just not the running one.

pradyunsg:
Hmm... Okie! I'm inclined to suggest that you maintain a tiny Python script that behaves exactly like you want it to, using the installer API -- for now, at the very least.

There's some discussion of an `--interpreter` option that behaves differently from what you're asking for here in https://github.com/pypa/installer/issues/98#issuecomment-1042484169; and I'm wary of just adding this as a CLI-maps-directly-to-API approach since one of the things I want to avoid is redistributors inventing their own target schemes instead of explicitly using what the interpreter would.

FFY00:
We would need to run the target interpreter to instrospect the install paths anyway, so if that would be viable, so should you running `installer` on that interpreter directly. As I pointed out in other issues, using the paths from the current interpreter and trying to modify them to fit what we would usually expect to be in an external target interpreter is a bad idea and will cause lots of headaches. So, as @pradyunsg suggests, I think you would be much better off maintaining your own little script that places the files in the locations you expect for your particular usage.

rossburton:
Our "host" Python is of the same version as the target and knows about target paths, so it mostly does do the right thing.  It's literally just the interpreter which needs updating.

rossburton:
I had a quick look at this. I don't understand how writing a custom wrapper to call the installer API will solve my problem, when it's the installer API writing the hashbangs in `scripts.py`.  Am I missing something?

pradyunsg:
[...]

PR #258, opened 2025-01-15
Implement `--executable' option (#257)

riastradh:
Resolves: #257

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -74,6 +74,28 @@ def test_main_prefix(fancy_wheel, tmp_path):
     }
 
 
+def test_main_executable(fancy_wheel, tmp_path):
+    destdir = tmp_path / "dest"
+
+    main(
+        [
+            str(fancy_wheel),
+            "-d",
+            str(destdir),
+            "--executable",
+            "/monty/python3.x",
+        ],
+        "python -m installer",
+    )
+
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -57,7 +57,18 @@ def test_main_multiple_wheels(fancy_wheel, another_fancy_wheel, tmp_path):
 def test_main_prefix(fancy_wheel, tmp_path):
     destdir = tmp_path / "dest"
 
-    main([str(fancy_wheel), "-d", str(destdir), "-p", "/foo"], "python -m installer")
+    main(
```
Secrus, line 60: I don't like hijacking the prefix test for that option. Please move your checks to a separate test.

Secrus (approved):

pradyunsg:
@Secrus This was something that had been discussed in https://github.com/pypa/installer/issues/107 prior, with the conclusion being to not do this. The specific request in #257 is effectively unchanged relative to that and I strongly think this is a misfeature we should not support.

Secrus:
@pradyunsg Oh, I didn't know that existed before... I will revert this feature. 

@riastradh sorry, I missed some history lessons on the previous decisions in the project. Thank you for doing the work, but your feature won't be happening.

Issue #260, opened 2025-01-19
relpath error with differing path roots


BrianPugh:
Extending conversation from https://github.com/python-poetry/poetry/issues/10028

With Poetry 2.0, on windows Github Action, there's an unhandled exception when " path is on mount 'C:', start on mount 'D:'". I'm not sure if this is a user (poetry) issue or an `installer` issue.

Full stack trace here, abbreviated to the important bits below:

```
  5  ~\poetry\venv\lib\site-packages\poetry\installation\wheel_installer.py:109 in install
      107|             )
      108| 
    > 109|             install(
      110|                 source=source,
      111|                 destination=destination,

  4  ~\poetry\venv\lib\site-packages\installer\_core.py:131 in install
      129| 
      130|     written_records.append((root_scheme, RecordEntry(record_file_path, None, None)))
    > 131|     destination.finalize_installation(
      132|         scheme=root_scheme,
      133|         record_file_path=record_file_path,

  3  ~\poetry\venv\lib\site-packages\installer\destinations.py:278 in finalize_installation
      276| 
      277|         record_list = list(records)
    > 278|         with construct_record_file(record_list, prefix_for_scheme) as record_stream:
      279|             self.write_to_fs(
      280|                 scheme, record_file_path, record_stream, is_executable=False

  2  ~\poetry\venv\lib\site-packages\installer\utils.py:211 in construct_record_file
      209|     writer = csv.writer(stream, delimiter=",", quotechar='"', lineterminator="\n")
      210|     for scheme, record in records:
    > 211|         writer.writerow(record.to_row(prefix_for_scheme(scheme)))
      212|     stream.seek(0)
      213|     return stream.detach()

  1  ~\poetry\venv\lib\site-packages\installer\destinations.py:271 in prefix_for_scheme
      269|             if file_scheme == scheme:
      270|                 return None
    > 271|             path = os.path.relpath(
      272|                 self.scheme_dict[file_scheme],
      273|                 start=self.scheme_dict[scheme],

  ValueError

  path is on mount 'C:', start on mount 'D:'

  at C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\ntpath.py:703 in relpath
      699|         path_abs = abspath(normpath(path))
      700|         start_drive, start_rest = splitdrive(start_abs)
      701|         path_drive, path_rest = splitdrive(path_abs)
      702|         if normcase(start_drive) != normcase(path_drive):
[...]

PR #272, opened 2025-04-03
Fix potential relpath bug on Windows platform

Secrus:
Resolves: #260


pradyunsg:
CI suggests that this fix isn't complete, and I think we should add a dedicated test covering the edge case we wanna cater to here.

PR #276, opened 2025-04-29
Fix bug when writing out scripts if the python interpreter has a space

eli-schwartz:
In SchemeDirectoryDestination, we write:

- files, which may be part of the wheel "scripts" scheme and if so must have #!python replaced by the interpreter shebang

- entrypoints, which are built from scratch with the interpreter shebang

Only in the latter case did we ensure that we build a portable shebang.

Partially reverts commit 0edcaa769d9caf965032907151c513c46154e70f, which refactored build_shebang into a class.

### DIFF
--- a/src/installer/scripts.py
+++ b/src/installer/scripts.py
@@ -43,6 +43,41 @@ if __name__ == "__main__":
 """
 
 
+def _is_executable_simple(executable: bytes) -> bool:
+    if b" " in executable:
+        return False
+    shebang_length = len(executable) + 3  # Prefix #! and newline after.
+    # According to distlib, Darwin can handle up to 512 characters. But I want
+    # to avoid platform sniffing to make this as platform-agnostic as possible.
+    # The "complex" script isn't that bad anyway.
+    return shebang_length <= 127
+
+
+def _build_shebang(executable: str, forlauncher: bool) -> bytes:
+    """Build a shebang line.
+
+    The non-launcher cases are taken directly from distlib's implementation,
+    which tries its best to account for command length, spaces in path, etc.
+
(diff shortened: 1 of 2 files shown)


eli-schwartz:
This is not strictly a followup on #200 but it is something I happened to notice as an aside while I was working on that. Apologies for the lateness.

src/installer/scripts.py
```diff
@@ -43,6 +43,41 @@
...
+    The non-launcher cases are taken directly from distlib's implementation,
+    which tries its best to account for command length, spaces in path, etc.
+
+    https://bitbucket.org/pypa/distlib/src/58cd5c6/distlib/scripts.py#lines-124
+    """
+    executable_bytes = executable.encode("utf-8")
+    if forlauncher:  # The launcher can just use the command as-is.
+        return b"#!" + executable_bytes
+    if _is_executable_simple(executable_bytes):
+        return b"#!" + executable_bytes
```
Secrus, line 68: I have just noticed, that for both those cases we return exactly the same value. Can you make it into a single `if` statement with an `or`?

src/installer/utils.py
```diff
@@ -24,6 +24,8 @@
     cast,
 )
 
+from installer.scripts import _build_shebang
```
Secrus, line 27: Not sure that I like importing and using a "private" function. I think a better solution would be to extract the part of the logic that is required for this use case into another function and having the `_build_shebang` use that. That should also allow `_build_shebang` to remain a part of the class, as it was before.

Secrus (changes requested):

src/installer/scripts.py
eli-schwartz: I have a followup commit that does this. It's not really appropriate when moving a block of code from one location to another, to also start changing the interior design of that block.

PR #294, opened 2025-07-31
PEP 639 compliance

DimitriPapadopoulos:
Keep the redundant Trove classifier for now, until tools such as pip-licenses support PEP 639.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 build-backend = "flit_core.buildapi"
 requires = [
-    "flit_core<4,>=3.2",
+    "flit_core<4,>=3.11",
 ]
 
 [project]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,8 +11,9 @@ authors = [
     { name = "Pradyun Gedam", email = "pradyunsg@gmail.com" },
 ]
 requires-python = ">=3.9"
+license = "MIT"
 classifiers = [
-    "License :: OSI Approved :: MIT License",
+    "License :: OSI Approved :: MIT License",  # for compatibility with tooling such as pip-licenses
     "Programming Language :: Python :: 3 :: Only",
     "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> but doesn't support PEP 639 - created more than 5 years ago.

This is an extremely deceptive statement and should be retracted.

DimitriPapadopoulos:
> > but doesn't support PEP 639 - created more than 5 years ago.
> 
> This is an extremely deceptive statement and should be retracted.

Why deceptive? It's a real noob question, not trolling. Please bear in mind, I'm no specialist, but only an end-user that finds the constant changes in the build process in the last 5 to 10 years worse than the Python 2 to 3 transition. It's impossible to reverse the tide by now, at least I'd like some consistency in the ecosystem.

DimitriPapadopoulos:
Upstream issue: https://github.com/pypa/flit/issues/753.

eli-schwartz:
It was finally approved on August 2024, just under a year ago -- it also spent five years before that ***being argued about***, so expecting tools to have implemented it six years ago just doesn't work.

Especially considering the python ecosystem has an extreme preference to avoid implementing anything in the standards arena that isn't a final "done deal, signed and sealed". People aware of PEP 639 still specifically refused to implement it because it wasn't approved and they didn't want to get scolded for violating the standards by implementing something that was only a proposal.

The price we pay for this ecosystem perspective is that the implementation of standards is artificially slowed down and rollout takes many years. I'd say that this then obligates build backends to refrain from emitting fatal errors if projects want to support both the old and new way.

pyproject.toml
```diff
@@ -11,8 +11,8 @@ authors = [
     { name = "Pradyun Gedam", email = "pradyunsg@gmail.com" },
 ]
 requires-python = ">=3.9"
+license = "MIT"
```
Secrus, line 14: You could also add `license-files` key here.

DimitriPapadopoulos: To 3.11, yes. I keep forgetting that!

PR #328, opened 2026-04-04
Return install record mismatches from install()

dimbleby:
have `install()` return a list of (source_record, installed_record) tuples for files whose hash or size differs from the wheel's `RECORD`.  Callers that don't care can ignore the return value (previously None).

Alternative approach to #327, allowing installers to detect (most) errors in only a single pass over large files.  They would
- `validate_record(validate_contents=False)`
- `install()` and pay attention to reported mismatches.

This is not perfect
- `RECORD` that is wrong for scripts gets missed, because installer rewrites scripts during install
  -  callers who want to be more careful can still call `validate_record()`
- `RECORD` that uses an algorithm other than one of the blessed few is not checked
  - again `validate_record()` is still available.
- I do not love the manipulation of `Destination.hash_algorithm`, but it would be a breaking change to pass this through more cleanly as a parameter.  
  - You have just gone 1.0, I speculate that you might feel it is too soon for a 2.0

But it covers most things in an efficient way.

Maintainers will need to take a view on whether this is a reasonable set of trade-offs or not.


Secrus:
@pradyunsg I would like your input on that

PR #330, opened 2026-04-08
fix: preserve original paths when using destdir with symlinked schemes

Bahtya:
## Problem

When using `destdir` with symlinked scheme paths (e.g. `/usr/bin/pyproject-build` → `/usr/bin/pyproject-build-3.13`), `_path_with_destdir` resolves symlinks before computing the destdir-relative path. This causes the resolved name (`pyproject-build-3.13`) to be used instead of the original (`pyproject-build`).

## Root Cause

`resolve()` follows symlinks on both `target_dir` and `file` (lines 140-141). When `destdir` is set, `file.relative_to(file.anchor)` operates on the resolved path, losing the original filename.

## Solution

Use unresolved paths for destdir path computation. Only use `resolve()` for the safety check that ensures files stay within the target directory.

```python
# Before:
target_dir = Path(self.scheme_dict[scheme]).resolve()
file = (target_dir / path).resolve()

# After:
target_dir = Path(self.scheme_dict[scheme])
file = target_dir / path
resolved_dir = target_dir.resolve()
resolved_file = file.resolve()
# Use resolved_* for safety check, file (unresolved) for destdir
```

Fixes #325

### DIFF
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -155,7 +155,10 @@ class SchemeDictionaryDestination(WheelDestination):
             )
 
         if self.destdir is not None:
-            rel_path = file.relative_to(file.anchor)
+            # When using destdir, construct the path from the original (non-absolute)
+            # scheme paths so that symlinked scheme directories are preserved correctly.
+            orig_dir = Path(self.scheme_dict[scheme])
+            rel_path = (orig_dir / path).relative_to(orig_dir.anchor)
             return Path(self.destdir) / rel_path
         return file
 


jdknight:
This change fixes my tooling issue. Although, the change now needs a tweak based on the changes outlined in https://github.com/pypa/installer/pull/326.

Bahtya:
Thanks for the heads-up! I've rebased on main and adapted the fix to work with the new `os.path.abspath()` approach from #326.

The conflict resolution was straightforward:
- Keep the upstream `os.path.abspath()` for path safety checks (no `resolve()`, no `is_relative_to` helper)
- Fix the destdir path construction to use the original (non-absolute) scheme paths, so symlinked scheme directories are preserved correctly

All 150 tests pass locally. CI should be green shortly.

pradyunsg:
With https://github.com/pypa/installer/pull/326/, this should no longer be necessary.

Bahtya:
Thanks @pradyunsg! Since #326 has been merged and addresses the underlying issue, this PR is no longer necessary. Closing.

jdknight:
Note that #326 does not resolve the issue outlined in https://github.com/pypa/installer/issues/325#issuecomment-4192973785, which this PR indirectly fixed.

But maybe it was wrong of me to add that comment in there, focusing on it as an issue with changes to `_path_with_destdir` versus making a separate bug for loss in a relative scheme path for installs.

PR #175, opened 2023-03-23
use `WheelDestination.interpreter` to compile

BlueGlassBlock:
Resolves #172

I'm thinking about how to reach 100% test.

PR #332, opened 2026-05-15
installer: add --interpreter option

rossburton:
The currently running Python might not be the right one for the installed wheels, so add an option to allow the interpreter to be overridden.

This is essential in cross/distro builds where we want to install a wheel into a working directory to build a distro package (eg a rpm or deb) with pypa/installer, but the currently running python on the build host might not be the right path for the scripts to use on the target.

### DIFF
--- a/src/installer/__main__.py
+++ b/src/installer/__main__.py
@@ -31,6 +31,13 @@ def _get_main_parser() -> argparse.ArgumentParser:
         help="override prefix to install packages to",
     )
     parser.add_argument(
+        "--interpreter",
+        "-i",
+        type=str,
+        default=sys.executable,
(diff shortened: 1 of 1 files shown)


Secrus:
Previously rejected with #258 #263 #107

rossburton:
I'd actually forgotten that I'd been here before in #107.

Does your position still stand that everyone who is currently patching these few lines in need to continue patching forever? The previous statement was basically just write your own installer tool that is 99% identical to pypa/installer but with one extra option, which seems like pointless duplication.

Especially as at the time the argument was "the API supports that", but #175 merging absolutely means that it would not anymore.

PR #339, opened 2026-06-09
Skip script collisions with entry points

Himanshuagrawal4:
## Summary
- track entry point scripts written during installation
- skip `.data/scripts` files that would overwrite an already generated entry point script
- emit a runtime warning for the skipped conflicting script
- add a regression test for a wheel containing both forms with the same script name

Closes #315.

## Validation
- `PYTHONPATH=src python -m pytest tests/test_core.py`
- `python -m compileall -q src\installer`
- git diff --check

### DIFF
--- a/src/installer/_core.py
+++ b/src/installer/_core.py
@@ -81,6 +81,7 @@ def install(
     # RECORD handling
     record_file_path = posixpath.join(source.dist_info_dir, "RECORD")
     written_records = []
+    generated_script_paths: set[str] = set()
 
     # Write the entry_points based scripts.
     if "entry_points.txt" in source.dist_info_filenames:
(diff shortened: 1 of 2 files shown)


eli-schwartz:
"In the face of ambiguity, refuse the temptation to guess."

Himanshuagrawal4:
Thanks. I agree this is a policy choice. This PR follows one of the options from #315 and pip's current behavior: when a legacy script and generated entry point target the same script path, keep the generated entry point and skip the legacy script. Unlike pip, it emits a RuntimeWarning so callers can see that the wheel is ambiguous. Happy to change direction if maintainers prefer failing with a clearer error or exposing a policy knob instead.

PR #340, opened 2026-06-18
fix: replace deprecated posixpath.commonprefix() with startswith()

frenzymadness:
os.path.commonprefix() is deprecated in Python 3.15 and emits a DeprecationWarning. Replace all three usages with str.startswith() checks, which are also semantically more correct (commonprefix does character-level matching and can produce false positives for paths sharing a string prefix but not a directory boundary).

Without this, `TestInstall.test_skips_pycache_and_warns` fails with Python 3.15 because the code generated more warnings than expected.

### DIFF
--- a/src/installer/_core.py
+++ b/src/installer/_core.py
@@ -42,7 +42,7 @@ def _determine_scheme(
     data_dir = source.data_dir
 
     # If it's in not `{distribution}-{version}.data`, then it's in root_scheme.
-    if posixpath.commonprefix([data_dir, path]) != data_dir:
+    if not path.startswith(data_dir + "/"):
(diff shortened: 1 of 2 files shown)


Secrus:
Python docs mention [using `commonpath` instead of `commonprefix`](https://docs.python.org/3.15/library/os.path.html#os.path.commonprefix) yet you used `str.startswith`. Any particular reason for that?

frenzymadness:
Nothing special, startswith was just the first thing that crossed my mind. `startswith` is also slightly defensive because commonpath raises ValueError when absolute and relative paths are mixed.

Secrus:
Superseded by #345

PR #342, opened 2026-06-19
perf: avoid building a Path object for each record entry

radoering:
Splitting record entry paths by `"/"` should be safe because path separators are normalized in `parse_record_file`.


faulknerdeana111-ops (approved):

Secrus (approved):

faulknerdeana111-ops:
Thank you

PR #343, opened 2026-06-24
Recognize build tags in wheel data directories

### LINKED ISSUE #193: `data_dir` lacks optional build tag
This is related to #171 but this issue impacts all wheels with a build tag.

For example, using the following command to create a wheel for a package named `foo` with version 0.4.2:

```bash
python setup.py bdist_wheel --build-number 1337
```

will create a wheel named `foo-0.4.2-1337-py3-none-any.whl`. The `build_tag` in:

https://github.com/pypa/installer/blob/444e5295f2403ab303dcf14160591074396f1d21/src/installer/utils.py#L66-L68

will be `1337`.

Data files in the wheel will have paths such as `foo-0.4.2-1337.data/some_folder/some_file.txt`. Due to the implementation of `data_dir`:

https://github.com/pypa/installer/blob/444e5295f2403ab303dcf14160591074396f1d21/src/installer/sources.py#L49-L52

 we'll thus have the following variable values here:

https://github.com/pypa/installer/blob/444e5295f2403ab303dcf14160591074396f1d21/src/installer/_core.py#L43

```python
assert data_dir == "foo-0.4.2.data"
assert path == "foo-0.4.2-1337.data/some_folder/some_file.txt"
assert posixpath.commonprefix([data_dir, path]) == "foo-0.4.2"
assert posixpath.commonprefix([data_dir, path]) != data_dir
# -> the scheme is incorrectly determined as root_scheme.
```

lin-hongkuan:
Fixes #193.

This updates `WheelFile.data_dir` to include the optional wheel build tag parsed from the wheel filename, so files under directories such as `foo-0.4.2-1337.data/` are routed through the wheel `.data` scheme instead of being treated as root-scheme files.

The base `WheelSource` behavior is unchanged for callers that construct sources without a wheel filename.

Tests added:
- `WheelFile.data_dir` includes the build tag for build-tagged wheels.
- `install()` routes a build-tagged wheel data file to the `data` scheme.

Checks run locally:
- `PYTHONPATH=src python -m pytest tests\test_sources.py::TestWheelFile::test_build_tag_is_part_of_data_dir tests\test_core.py::TestInstall::test_handles_wheel_data_dir_with_build_tag -q`
- `PYTHONPATH=src python -m pytest tests\test_sources.py tests\test_core.py -q`
- `PYTHONPATH=src python -m pytest -q`
- `python -m ruff check src tests`
- `python -m ruff format --check src tests`
- `PYTHONPATH=src python -m mypy src`
- `git diff --check`

### DIFF
--- a/src/installer/sources.py
+++ b/src/installer/sources.py
@@ -167,6 +167,7 @@ class WheelFile(WheelSource):
 
         basename = Path(f.filename).name
         parsed_name = parse_wheel_filename(basename)
+        self._build_tag = parsed_name.build_tag
         super().__init__(
             version=parsed_name.version,
             distribution=parsed_name.distribution,
@@ -211,6 +212,14 @@ class WheelFile(WheelSource):
             )
 
         return dist_info_dir
+
+    @cached_property
+    def data_dir(self) -> str:
+        """Name of the data directory."""
+        if self._build_tag is None:
+            return super().data_dir
+
+        return f"{self.distribution}-{self.version}-{self._build_tag}.data"
 
     @property
     def dist_info_filenames(self) -> list[str]:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,12 +10,19 @@ def fancy_wheel(tmp_path):
 
 
 @pytest.fixture
+def fancy_wheel_with_build_tag(tmp_path):
+    return mock_wheel(tmp_path, "fancy", build_tag="1337")
+
+
+@pytest.fixture
 def another_fancy_wheel(tmp_path):
     return mock_wheel(tmp_path, "another_fancy")
 
 
-def mock_wheel(tmp_path, name):
-    path = tmp_path / f"{name}-1.0.0-py2.py3-none-any.whl"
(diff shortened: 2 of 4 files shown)


faulknerdeana111-ops (approved):

yangfan-yf-yf (changes requested):
﻿The optional build tag belongs only in the wheel filename. The wheel specification defines the data directory as the dist-info directory name with the extension changed to `.data`, so a build-tagged wheel still uses `distribution-version.data`.

Changing `WheelFile.data_dir` to include `self._build_tag` makes a compliant wheel such as `fancy-1.0.0-1337-py3-none-any.whl` with `fancy-1.0.0.data/data/...` no longer match the data scheme. The new fixture and assertions instead encode the non-standard `fancy-1.0.0-1337.data` layout.

Could this preserve the required `distribution-version.data` handling for build-tagged wheels? If compatibility with legacy non-standard artifacts is desired, it needs to be additive rather than replacing the standard directory name, with regression coverage for both layouts.

PR #345, opened 2026-06-26
Use os.path.commonpath over commonprefix

ichard26:
`commonprefix` was deprecated in Python 3.15 as it does a character-by-character check which is 99% likely to be not what you actually want.

`commonpath` does the sane thing of returning only valid paths (aka respecting the directory structure).

Having briefly read the code, I assume that there is no mixing of absolute and relative paths at these callsites. If there is, then `commonpath` won't work. 

This was caught in pip's test suite on Python 3.15.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Use `os.path.commonpath` instead of `os.path.commonprefix` to address deprecation
+  warnings on Python 3.15 (#345)
+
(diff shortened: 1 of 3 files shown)


ichard26:
Oh, I just noticed there is already an open PR for this: https://github.com/pypa/installer/pull/340. Feel free to close this in favour of that.

(I did not use AI to write this PR, FWIW, haha.)

Secrus (approved):

PR #347, opened 2026-06-28
Silence bytecode compile failures during install

itscloud0:
## Summary
- silence bytecode compilation failures by using `compileall`'s fully quiet mode
- keep installer behavior aligned with pip by still ignoring compile failures instead of aborting installation
- add a regression test that locks in silent handling for failed bytecode compilation

Closes #311

## Problem
`installer` currently ignores bytecode compilation failures, but it still emits `*** Error compiling ...` messages to stderr. That is noisy for wheels that intentionally contain Python files targeting newer syntax, and it does not match pip's behavior noted in the issue discussion.

## Testing
- `uv run --python 3.10 --with pytest pytest tests/test_main.py -q`
- `uv run --python 3.10 --with pytest --with pytest-cov --with pytest-xdist pytest -q`
- `uv run --python 3.10 --with ruff ruff check src/installer/destinations.py tests/test_main.py`
- `uv run --python 3.10 --with ruff ruff format --check src/installer/destinations.py tests/test_main.py`

## Notes
- I did not change install success/failure semantics; this only removes the noisy stderr output while preserving the existing ignore-and-continue behavior discussed on the issue.

### DIFF
--- a/src/installer/destinations.py
+++ b/src/installer/destinations.py
@@ -268,7 +268,7 @@ class SchemeDictionaryDestination(WheelDestination):
 
         for level in self.bytecode_optimization_levels:
             compileall.compile_file(
-                target_path, optimize=level, quiet=1, ddir=dir_path_to_embed
+                target_path, optimize=level, quiet=2, ddir=dir_path_to_embed
             )
 
     def finalize_installation(
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,4 +1,6 @@
 import os
+import sys
+import types
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


yangfan-yf-yf (approved):
﻿Reviewed `249a4e3` independently.

This uses `compileall`'s documented fully quiet mode while preserving the existing non-fatal compilation behavior discussed in #311. The regression test verifies that the installation path invokes compilation with `quiet=2` and does not emit stderr output.

Validation on Windows with Python 3.12:

- `python -m pytest -q` ? 151 passed
- `python -m mypy src` ? passed
- `git diff --check` ? clean

The GitHub cross-platform test matrix and pre-commit check are passing.

PR #354, opened 2026-09-02
Read RECORD with the default csv reader, not str.splitlines()

dylanpulver:
`construct_record_file` writes RECORD with `csv.writer`, so a path containing a newline comes out correctly quoted. Both read sites — `WheelFile.validate_record` and `WheelFile.get_contents` — called `str.splitlines()` before handing the result to `parse_record_file`, so installer cannot read the RECORD it just wrote.

Measured on `59f0a65`, using installer's own writer and then its own reader:

```
wrote  '"demo/we\nird.py",sha256=...,6'
read   'demo/weird.py'           (the newline is gone)

validate_record() on that wheel:
  In demo-1.0-py3-none-any.whl, demo/we
  ird.py is not mentioned in RECORD
```

`splitlines()` also splits on characters `csv` does not treat as row separators. I ran the writer and then the reader for all ten: `\n` and `\r` come back with the character silently deleted; `\v`, `\f`, `\x1c`, `\x1d`, `\x1e`, `\x85`, U+2028 and U+2029 raise `expected 3 elements, got 1`. All ten read back byte-identical through `csv.reader` over the same file.

The recording-installed-projects spec pins the dialect: RECORD "must be readable with the default `reader` of Python's `csv` module", line terminator "either `\r\n` or `\n`". Interposing `splitlines()` is a different, larger set of terminators.

Fix is to hand the RECORD text to `parse_record_file` as `io.StringIO(text, newline="")`.

Ran `pytest tests/`: 152 passed, `coverage report` still 100%. Reverting only `src/installer/sources.py` to `origin/main` fails both new cases (150 passed, 2 failed). `splitlines(keepends=True)` is the obvious narrower fix and **it does pass the newline case** — csv reassembles the quoted field from the kept terminators — but it still fails the U+2028 case, which is why the test parametrises both. `ruff check` reports the same 3 pre-existing `PLR1704` on an unmodified checkout and none in what I touched.

I did not change `parse_record_file` itself beyond one docstring line; it already accepts any iterable of strings.

AI assistance: this change was drafted with Claude Opus 5 (`claude-opus-5`). The outputs above were produced by running installer's own writer and reader locally against this branch.

### DIFF
--- a/src/installer/records.py
+++ b/src/installer/records.py
@@ -222,7 +222,8 @@ def parse_record_file(rows: Iterable[str]) -> Iterator[tuple[str, str, str]]:
     Returns an iterable of 3-value tuples, that can be passed to
     :any:`RecordEntry.from_elements`.
 
-    :param rows: iterator providing lines of a RECORD (no trailing newlines).
+    :param rows: iterator providing lines of a RECORD (no trailing newlines),
+        or a text stream opened with ``newline=""``.
     """
     reader = csv.reader(rows, delimiter=",", quotechar='"', lineterminator="\n")
     for row_index, elements in enumerate(reader):
--- a/src/installer/sources.py
+++ b/src/installer/sources.py
@@ -1,5 +1,6 @@
 """Source of information about a wheel file."""
 
+import io
 import posixpath
 import stat
 import zipfile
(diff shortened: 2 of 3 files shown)


PNHD (approved):
Reviewed exact head `c9373f2e378f605c0e9559805b8a950eacdc2ba7`. SECURITY-N/A.

The change matches both authoritative contracts that matter here: the PyPA RECORD specification requires the CSV dialect to be readable by Python's default `csv.reader`, and Python's `csv` documentation says CSV text streams should use `newline=""` so the csv module can handle embedded newlines itself. Replacing `str.splitlines()` with `io.StringIO(record_text, newline="")` at both RECORD read sites preserves that boundary.

The regression covers both an embedded newline inside a quoted path and U+2028, which `str.splitlines()` treats as a separator even though the RECORD dialect does not. Current CI is green, there are no existing PNHD reviews or maintainer objections, and I found no remaining correctness, compatibility, readability, architecture, security, or performance blocker on this head. LGTM.

Secrus (commented):
While I get the reasoning of this patch, current behavior matches `pip` and we don't want to diverge from what `pip` does, since that is still the de facto standard for the package installers.

PR #355, opened 2026-09-21
Update GH Actions pins and add zizmor

Secrus:


.github/workflows/ci.yml
```diff
@@ -42,10 +46,12 @@ jobs:
       # prettier-ignore
       - run: >
           nox
-          -s test-${{ matrix.python-version }}
-          doctest-${{ matrix.python-version }}
+          -s test-${PYTHON_VERSION}
+          doctest-${PYTHON_VERSION}
           --error-on-missing-interpreters
         if: matrix.python-version != 'pypy3.10'
```
hugovk, line 52:
```suggestion
        shell: bash
        if: matrix.python-version != 'pypy3.10'
```

Secrus: That worked, thanks :smile:

.pre-commit-config.yaml
```diff
@@ -30,3 +30,9 @@ repos:
     hooks:
       - id: ruff-check
       - id: ruff-format
+
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
+    rev: v1.30.1
+    hooks:
+      - id: zizmor
+        args: ["-p"]
```
hugovk: My personal preference is to put the longer versions in scripts, because it's hard to remember what short opts mean (beyond the super common `ls -l` and friends)

```suggestion
        args: ["--pedantic"]
```

hugovk (approved):
Welcome to the zizmor club! 🔒

.pre-commit-config.yaml
Secrus: Good point

